PR #8, opened 2019-12-24
Update rxdart 0.22.6 -> 0.23.1

BeMacized:
Allows `just_audio` 0.0.2 to be used with `audio_service` 0.5.5.


yringler:
~I think you want to run flutter pub get to update pubspec.lock~
just_audio already uses 0.23.1, and has that in the lock! Yeah, this is a good idea.

BeMacized:
Looks like this is now in the current master.

PR #11, opened 2019-12-28
Allow "playbackState" to be called before setting a URL.

BeMacized:
Currently, when calling `playbackState` on an `AudioPlayer` instance before `setUrl` has been called at least once, a `NoSuchMethodError` is thrown. I would expect either `null` or `AudioPlaybackState.none` to be returned.

Test:
```dart
AudioPlayer player = AudioPlayer();
print(player.playbackState); // NoSuchMethodError
await player.setUrl("https://s3.amazonaws.com/scifri-episodes/scifri20181123-episode.mp3");
await player.play();
```
Stacktrace: 
```
E/flutter ( 5492): [ERROR:flutter/lib/ui/ui_dart_state.cc(157)] Unhandled Exception: NoSuchMethodError: The getter 'state' was called on null.
E/flutter ( 5492): Receiver: null
E/flutter ( 5492): Tried calling: state
E/flutter ( 5492): #0      Object.noSuchMethod (dart:core-patch/object_patch.dart:53:5)
E/flutter ( 5492): #1      AudioPlayer.playbackState (package:just_audio/just_audio.dart:111:63)
```

This patch has it return `AudioPlaybackState.none` if no value is yet present.

### DIFF
--- a/lib/just_audio.dart
+++ b/lib/just_audio.dart
@@ -108,7 +108,7 @@ class AudioPlayer {
       _playbackEventSubject.stream;
 
   /// The current [AudioPlaybackState].
-  AudioPlaybackState get playbackState => _audioPlaybackEvent.state;
+  AudioPlaybackState get playbackState => _audioPlaybackEvent?.state ?? AudioPlaybackState.none;
 
   /// A stream of [AudioPlaybackState]s.
   Stream<AudioPlaybackState> get playbackStateStream =>


ryanheise:
Sorry for the delay, I have been thinking about how I really want to handle this case. This pull request does solve the issue for `playbackState` but not `playbackStateStream` which should also be seeded with `none`. Probably what should happen is that the platform side should broadcast a playback event with a state of `none` upon instantiation which will propagate to all other variables and streams.

Although there will still be a brief moment before that first event is broadcast when your code may still be helpful for `playbackState` although I think the same should be done for `playbackEvent`.

volgin:
This is a good suggestion. There is no reason to delay this pull request - no plugin should have unhandled exceptions. No other solution is going to override the need to have a null check at this point.

ryanheise:
So ideally what I'd like to happen is that upon instantiation, there is an initial event containing a state change to `none`. This propagates everywhere ensuring that there is an initial item in each behavior subject, an initial event, and an initial state. The null exception would vanish because the fields are never null. But also, if your UI listens to any of the streams, they'll start working immediately and won't hold up your display.

This pull request has the right idea to initialise the state to `none`, although given the planned direction above, I would instead prefer to initialise the event to contain initial values including a state of `none`, which will propagate to the state getter that you've addressed.

Thanks for your patience, and I've committed the temporary fix. The rest of the plan will happen as part of #21 where I re-do the streams.

PR #24, opened 2020-01-20
Enable playing local files in iOs.

volgin:
There is no need to prefix a local file path with "file://". We just need to use 'fileURLWithPath:url'.

Tested with both .mp3 and .m4a files on iPhone 8 emulator and iPhone 11 device.

### DIFF
--- a/ios/Classes/AudioPlayer.m
+++ b/ios/Classes/AudioPlayer.m
@@ -155,7 +155,16 @@
 		[[NSNotificationCenter defaultCenter] removeObserver:_endObserver];
 		_endObserver = 0;
 	}
-	AVPlayerItem* playerItem = [[AVPlayerItem alloc] initWithURL:[NSURL URLWithString:url]];
+
+    AVPlayerItem *playerItem;
+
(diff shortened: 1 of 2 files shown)


ryanheise:
Thanks!

I did suspect I would need to change that to `fileURLWithPath:url` someday, although before writing that extra case in the code, I had my fingers crossed that `initWithURL` would would support file:// URLs. Thank you for testing it and reporting back that it does not work, I really appreciate that as it is not as easy for me to confirm such things through MacInCloud.

Now, on Android, the native `MediaPlayer` API does indeed accept file paths in addition to URIs, but ExoPlayer (which the plugin now uses) is documented as accepting URIs. I'm not sure if you have already tested this on Android, but even if it would work as a file path, I'd be a bit worried about this undocumented behaviour changing.

So I might suggest instead of changing the dart code to remove the 'file://' prefix, you could instead change the iOS code to check if `[url hasPrefix:@"file://"]` and if so, just take the substring after this as the file path. I guess `[url substringFromIndex:7]`.

PR #32, opened 2020-02-02
FLAC Support

Marekkon5:
### This PR:

Hello. This is a dirty fork with FLAC support. It works, but it's done the improper way.
I forked ExoPlayer, precompiled the FLAC Extension and pushed everything to GitHub (https://github.com/Marekkon5/ExoPlayer). Then I used JitPack to make the repo a gradle dependency (https://jitpack.io/#Marekkon5/ExoPlayer). I've replaced the Google's ExoPlayer dependency with the JitPack one - see changes in:  ` android/build.gradle`. 

### My other attemps at making FLAC work in `just_audio`:

I've also tried doing it the "proper" way (by using local install of ExoPlayer) see bottom of https://github.com/google/ExoPlayer/blob/release-v2/README.md and compiling the FLAC extension: https://github.com/google/ExoPlayer/blob/release-v2/extensions/flac/README.md 
It works, but -  `just_audio/android/settings.gradle` is completly ignored. You have to use `just_audio/example/android/settings.gradle` or your app's `android/settings.gradle` and add the lines from ExoPlayer README bottom. It is bad solution, because everyone would have to add those lines in their app, and you can't compile the app for release.

Thank you.


ryanheise:
This is a great contribution, thanks! I want to make sure I understand a few things. First, in the FLAC readme it says:

> License note
> 
> Please note that whilst the code in this repository is licensed under Apache 2.0, using this extension also requires building and including one or more external libraries as described below. These are licensed separately.

I didn't see the licenses of those external libraries mentioned on that page. Do you know what the licenses are? Flutter's license registry will not pick these up automatically so we'll need to include them in the license file manually.

Regarding the proper vs dirty way, do you know if there's a way to publish just the extension via github/jitpack and be able to add the extension to gradle in combination with the official exoplayer dependencies?

Marekkon5:
Hello. I will research the licensing later.

To the official ExoPlayer + the fork - I tried that but got errors (it was duplicate class error I think) so I replaced the official with the forked. Maybe cleaning the dependency caches or cleaning the project will work. Will also do research later.

Marekkon5:
### FLAC Licensing:
I've used ExoPlayer's guide to build FLAC (https://github.com/google/ExoPlayer/tree/release-v2/extensions/flac). It uses FLAC Source Code from: https://ftp.osuosl.org/pub/xiph/releases/flac/flac-1.3.2.tar.xz
The Licenses are included in that archive, and also in my prebuilt repo under `ExoPlayer/extensions/flac/src/main/jni/flac/` 
All the licenses:
README: https://github.com/Marekkon5/ExoPlayer/blob/release-v2/extensions/flac/src/main/jni/flac/README
FDL: https://github.com/Marekkon5/ExoPlayer/blob/release-v2/extensions/flac/src/main/jni/flac/COPYING.FDL
GPL: https://github.com/Marekkon5/ExoPlayer/blob/release-v2/extensions/flac/src/main/jni/flac/COPYING.GPL
LGPL: https://github.com/Marekkon5/ExoPlayer/blob/release-v2/extensions/flac/src/main/jni/flac/COPYING.LGPL
Xiph: https://github.com/Marekkon5/ExoPlayer/blob/release-v2/extensions/flac/src/main/jni/flac/COPYING.Xiph

### Official + Fork
Unsuccessful Attempts:
1. My FLAC + rest original gives a lot of `Duplicate class` errors.
2. My FLAC fork + google's ExoPlayer using JitPack - same result as above

However, I got it working, by downloading the precompiled `aar` FLAC extension from my JitPack (https://jitpack.io/com/github/Marekkon5/ExoPlayer/extension-flac/release-v2-r2.11.1-gf1a2594-3/extension-flac-release-v2-r2.11.1-gf1a2594-3.aar)
And adding it as library in `build.gradle` alongside with official ExoPlayer's dependency links.
See commit: https://github.com/Marekkon5/just_audio/commit/facb6c8c6f89f61f6c7bfceee434290d6005e314 

Thank you.

ryanheise:
The licensing situation looks a bit complicated. I'm going to have to look into it further.

I downloaded the archive you linked to and its README says:

```
The libraries (libFLAC, libFLAC++) are
licensed under Xiph.org's BSD-like license (see COPYING.Xiph).  All other
programs and plugins are licensed under the GNU General Public License
(see COPYING.GPL).  The documentation is licensed under the GNU Free
Documentation License (see COPYING.FDL).
```

I won't be able to incorporate any GPL code since that license requires any app using the library to also be released in source code form under the GPL which would be unacceptable to most users of `just_audio`. The Xiph license is probably OK, but is this going to be functional without the GPL parts?

ryanheise:
By the way, this probably also explains why the FLAC extension was not included in ExoPlayer by default. If the GPL code is required to be functional, then we may have to consider other options.

One option is to figure out a way for users of the plugin to link in this extension without requiring everyone to use it.

Another option is to revert to `MediaPlayer` for FLAC audio.

ryanheise:
Looking at https://xiph.org/flac/license.html I'm a bit more optimistic:

> The reference implementation libraries are licensed under the New BSD License. In simple terms, these libraries may be used by any application, Open or proprietary, linked or incorporated in whole, so long as acknowledgement is made to Xiph.org Foundation when using the source code in whole or in derived works. The Xiph License is free enough that the libraries have been used in commercial products to implement FLAC, including in the firmware of hardware devices where other Open Source licenses can be problematic. In the source code these libraries are called libFLAC and libFLAC++.
>
> The rest of the software that the FLAC project provides is licensed under the GNU General Public License (GPL). This software includes various utilities for converting files to and from FLAC format, plugins for audio players, et cetera. In general, the GPL allows redistribution as long as derived works are also made available in source code form according to compatible terms.

That means that I can link to your build of ExoPlayer provided the binary doesn't include any of those components listed in the second paragraph above. Would you be able to confirm that that is the case?

Marekkon5:
I am not a license expert or anything, but if I understand correctly - the libFLAC library is under Xiph license (which is ok to use in this case) and the tools (`flac`, `metaflac`) are under GPL which is problem in this case?
If my assumptions are correct, then: ExoPlayer's flac extension is just JNI wrapper for libFLAC, so only the result libraries (`libflacJNI.so`) and ExoPlayer's java code are required. My ExoPlayer fork contains also FLAC source code, so I can add that to `.gitignore` and leave only the required files there (https://github.com/Marekkon5/ExoPlayer/tree/release-v2/extensions/flac/src/main/libs).
The `aar` file should contain only the result libraries.

Anyway - I don't know for sure if it matches the licenses. You can make it optional. In both cases (JitPack or aar file) you only have to edit the `build.gradle` and restart the app. You can add that to readme or as a comment to `build.gradle`.

Thank you.

Jei:
I've just noticed that ExoPlayer 2.11.2 includes https://github.com/google/ExoPlayer/issues/6406, which should allow for FLAC playback without external libraries. Since #78 updated ExoPlayer to 2.11.4, wouldn't it be worth a shot to try a solution using the comments in https://github.com/google/ExoPlayer/issues/6392 as a reference?

Issue #28, opened 2020-01-23
Exceptions when transitioning between playback states


volgin:
The comments in the plugin say: "It is legal to invoke this method only from the following states:"

Is this a target platform limitation? All platforms or only one/some?

In either case, the plugin can handle some of these transitions more gracefully. For example, if .stop() is called when a state is AudioPlaybackState.buffering, a plugin can chain "pause" and "stop" calls (for both or one of the platforms where this is an issue), instead of throwing an exception.

@ryanheise I would love to hear your thoughts on that before implementing this logic in my own code. My concern is that the farther away we go from the platform calls, the higher the probability of unpredictable behavior (e.g. stop() called from the app code when a status was AudioPlaybackState.playing, but by the time it reached the platform the status temporarily changed to AudioPlaybackState.buffering, resulting in a failed stop()).

ryanheise:
I generally try to use the preconditions imposed by the platform. That's the goal, but I haven't quite gotten there when it comes to the connecting and buffering states where the logic is a little more difficult. I'm with you in that the goal should be for these preconditions to be as few as possible. I'm not there yet, but that is the direction.

ryanheise:
The latest git commit should address this for Android.

I've also updated the state model on both Android and iOS so that the `buffering` state is treated independently of the other states. So now you can be buffering while playing, and you can be buffering while pausing, and these states are distinguishable. However, on iOS, the way I've implemented it, I can't really detect buffering while pausing anyway.

Also, I noticed a bug on Android where if you switched between play/pause or pause/play in the middle of seeking (i.e. while buffering), it would get confused about which state to return to after seeking. This bug goes away after changing the state model.

Let me know how that goes for you, and if it also doesn't break anything on iOS, I'll publish it.

volgin:
This sounds promising. I will try it ASAP. My immediate reaction is that from a UX perspective, the most important state is “I cannot play because I am buffering”. I hope the new update makes this primary use case possible.

ryanheise:
Answering @ReaganRealones from #33

> @ryanheise, okay sure have checked it out. Understood, that but the issue is still, while the player is connecting and the user cancels, it still continues to connect and thereafter plays. i.e calling `stop` while the player is in `connecting` state does not cancel playing music,

Hi @ReaganRealones , are you experiencing this issue on iOS or Android or both?

ReaganRealones:
Yeah I am experiencing this on Android

ReaganRealones:
[...]

PR #39, opened 2020-02-05
Why was this removed?

jonatadashi:
Idk why, but without this my app would literally not work


ryanheise:
Buffering is still there, but it's just a new API where the buffering state is orthogonal. The new state model is strictly more expressive than the old one. This is explained in https://github.com/ryanheise/just_audio/issues/28#issuecomment-582201270_ .

Issue #40, opened 2020-02-08
AVPlayerItemStatusFailed on iOS


codeeuser:
I am using the existing example but update the audio source to Flutter asset file. I have included the asset in pubspec.yaml
Device: iPhone 11 Pro Max Simulator
Flutter: v1.12.13+hotfix.7-stable

# To add assets to your application, add an assets section, like this:
  assets:
    - assets/audio/sound.mp3

@override
  void initState() {
    super.initState();
    _player = AudioPlayer();
    _player.setAsset('assets/audio/sound.mp3');
  }

The printed log is following. Noted that 'AVPlayerItemStatusFailed' is NSLog message.

Performing hot restart...                                               
Restarted application in 1,233ms.
AVPlayerItemStatusFailed

Camerash:
I have found the solution to this problem.
Apparently, `AVPlayer` used in iOS requires the audio file to have file suffix to play properly (i.e.: ".mp3")
However, the current `setAsset` function creates a temporary file without the suffix, causing it not to work on the iOS side.

I can create a PR later to fix the issue, cheers 👍

ryanheise:
@Camerash Thank you!

Camerash:
For now `_cacheFile` return a file path that does not account for the format suffix, and I don't won't to just hard code `".mp3"` for the file name.
I propose to make the cache file path a variable instead of getter to account for the format suffix when setting it in `setAsset()`. Does that sound go to you? @ryanheise

ryanheise:
Looks good. Now merged and included in release 0.1.1. Thanks!

Issue #35, opened 2020-02-05
iOS Testing help wanted


ryanheise:
All features mentioned in the README as "untested" for iOS are virtually all implemented, although just not 100% confirmed through testing.

If an untested feature works for you, I would be grateful if you could report that below and I will reflect that in the README.

codeeuser:
I am using iOS. I found sound from assets is not working.

ryanheise:
Thanks @codeeuser , if you encounter a bug, I would suggest opening a new issue / bug report, supplying information such as the error logs and example code used to reproduce the bug.

Can you confirm whether anything did work?

MickaelHrndz:
Hello,

Audio streaming from URL is not working on an **iPhone 4s** with **iOS 9.3.6**. Just_audio doesn't output any message. Audio_service is wrapped around it but also tried with an AudioPlayer by itself. I have the wifi turned on.

On an **iPhone 5 (iOS 10.3.1) simulator**, XCode raises these errors :
```
Undefined symbol: _FlutterMethodNotImplemented
Undefined symbol: _FlutterEndOfEventStream
Undefined symbol: _OBJC_CLASS_$_FlutterEventChannel
Undefined symbol: _OBJC_CLASS_$_FlutterCallbackCache
Undefined symbol: _OBJC_CLASS_$_FlutterEngine
Undefined symbol: _OBJC_CLASS_$_FlutterMethodChannel
```
There is also a number of warnings. Here is a screenshot  : https://imgur.com/a/PteK1Zx

On **iPhone 5s (iOS 10.3.1)** simulator (and iPhones above), it works fine. I couldn't try a 4s simulator.

viniciuszani:
Tested `read from file` from iOS with `mp3` files and it works as expected.
One disclaimer: the AVPlayer apparently can't play files without an extension (`.mp3`). Be sure to add it when saving files locally to be played.

Models where tests were conducted:
- iPhone 6s (iOS 13.3.1)
- iPhone 11 (OS 13.3.1) 
- iPhone X (OS 13.3.1)

suragch:
What is the status on this? The lack of confirmed iOS support is the one thing that is making me hesitate to switch to this library.

snaeji:
I have been using this player and testing for two types on iOS:

- Mp3 files from url (Works great)
- HLS live streams with 2 hours seeking window (Small issues)

[...]

PR #45, opened 2020-02-25
Add proper extension to asset media cache file for iOS asset playback

Camerash:
This PR adds proper extension to the asset media cache file, which fixes `AVPlayerItemStatusFailed` error when iOS clients tries to play an asset media. (Fixes issue #40)

Also, local file playback, as well as asset playback works well in iOS client. We can update the README once the PR is accepted and new version has been released. (Related to #35)

### DIFF
--- a/lib/just_audio.dart
+++ b/lib/just_audio.dart
@@ -83,6 +83,8 @@ class AudioPlayer {
 
   double _speed = 1.0;
 
+  File _cacheFile;
+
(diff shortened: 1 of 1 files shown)


ryanheise:
Thanks, @Camerash . I'm currently unable to test this due to a Dart bug breaking my dev environment, but it looks like you have taken the right approach in the PR.

Unfortunately I will have to wait for the Dart bug fix before testing and merging.

Issue #43, opened 2020-02-15
Load remote audio faster iOS using automaticallyWaitsToMinimizeStalling


YaarPatandarAA:
[automaticallyWaitsToMinimizeStalling](https://developer.apple.com/documentation/avfoundation/avplayer/1643482-automaticallywaitstominimizestal?language=objc)

A Boolean value that indicates whether the player should automatically delay playback in order to minimize stalling.

YaarPatandarAA:
Swift Example:
```
// Setup player
player = AVPlayer.init(playerItem: playerItem)
player.automaticallyWaitsToMinimizeStalling = false
```

Obj-C Example:
```
AVPlayer *player = [[AVPlayer alloc] initWithPlayerItem:playerItem];
player.automaticallyWaitsToMinimizeStalling = false;
```

Should be where the player item is init

ryanheise:
I would consider this as an iOS-specific option, with the current behaviour being the default.

YaarPatandarAA:
Yeah even if it is added as a option that would be great. It helps a lot with loading remote audio files quicker on iOS. 

I have used this to reduce some urls which take 10+ seconds to start playing down to 3 seconds to start playing the audio file.

Camerash:
I have made a PR that implements this feature. Can you try it out on my branch? @YaarPatandarAA

Camerash:
Notice that you need to call this method AFTER setting the url, or else the `AVPlayer` instance in iOS native will be null before you set this option, rendering it useless

ryanheise:
@Camerash Although I haven't used this method yet, I believe it should still be useful in the spirit of @YaarPatandarAA 's example code if you call it after calling `setUrl` but without awaiting the result of `setUrl`.

Another way to make the code more robust could be to allow this method to be called before `setUrl`. On the iOS side, define a new boolean field to store the set value of this flag, and conditionally, if `setUrl` has already completed (by inspecting the state), set the `automaticallyWaitsToMinimizeStalling` property on the player. Then in the iOS implementation of `setUrl`, if the flag has been set, then set the `automaticallyWaitsToMinimizeStalling` property on the player. That way, `automaticallyWaitsToMinimizeStalling` will either happen immediately if called after `setUrl` has completed, or it will just record it effectively as pending and will be processed on the next `setUrl`.

Camerash:
@ryanheise I have implemented your suggestion to make the feature more robust 🙏

YaarPatandarAA:
[...]

PR #49, opened 2020-03-04
implement automaticallyWaitsToMinimizeStalling on iOS

Camerash:
Implements #43

### DIFF
--- a/ios/Classes/AudioPlayer.m
+++ b/ios/Classes/AudioPlayer.m
@@ -17,6 +17,7 @@
 	BOOL _buffering;
 	id _endObserver;
 	id _timeObserver;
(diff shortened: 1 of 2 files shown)


ryanheise:
Thanks for the PR, it all seems correct, although when I tested it, I didn't notice any difference between setting this flag to false vs true. How about on your end? In any case, I have merged it as is. Will do a release soon.

Ghost-017:
Is possible to manually set the value of _automaticallyWaitsToMinimizeStalling to false on Android? I'm using this package to play an online radio. My problem is just_audio constantly stalls and resumes on the latest audio instead of where it stalled on Android. But on my web app, just_audio is smooth.

PR #54, opened 2020-03-09
Macos support

sachaarbonel:
Hi @ryanheise thank's for this great library! I don't know what was your plan for bringing support for macos but here it is. There are no good way to do this for now so it's just an ugly copy paste of the ios folders with some changes of the headers import and podspec. This this how sqflite did it and there are no best practices instructions from the flutter team so until then ...
Also, I can confirm the example works


ryanheise:
Thanks! I do not have a mac to be able to test this, but I would be happy to accept a PR. I would prefer the MacOS source files to "include" the corresponding iOS source files, with some conditional preprocessor code to handle the different header file requirements, if possible, to ease maintenance when new features are added. I just had a look at Flutter's official plugins but unfortunately they don't even do a copy and paste job, they've rewritten the MacOS implementation in Swift while their iOS implementation is in Objective C!

Incidentally, I assume you reran `flutter create` within the project, but it seems as a result there are now Kotlin and Swift files in addition to the Java and Objective C files which doesn't seem right, so you may need to take care to use options that are consistent with the original project.

sachaarbonel:
Sorry about the example I reverted my changes to the other platforms and keeped only the macos one. Yeah I agree it's not ideal. I tried to add this to podspec in macos (+deleting all the Classes files in macos + conditional imports in ios headers) 
```podspec
s.source           = { :path => '../ios' }
  s.source_files = '../ios/Classes/**/*'
  s.public_header_files = '../ios/Classes/**/*.h'
```
but I get a module not found error

ryanheise:
I've asked the Flutter team for best practices on this issue (https://github.com/flutter/flutter/issues/52285), so let's see what they say. Thanks for cleaning up the config, I'll take a look later today. If there is no better way to do it, we may just go with this approach for now, and manually copy the `*.m` files across after each code change.

stuartmorgan:
> I tried to add this to podspec in macos (+deleting all the Classes files in macos + conditional imports in ios headers)
> 
> ```ruby
> s.source           = { :path => '../ios' }
>   s.source_files = '../ios/Classes/**/*'
>   s.public_header_files = '../ios/Classes/**/*.h'
> ```

The reason this doesn't work is because those are invalid patterns; podspecs [are not allowed to refer to sources outside their directory](https://guides.cocoapods.org/syntax/podspec.html#group_file_patterns).

YaarPatandarAA:
Just tried this on an App I have for the Mac and it is working perfectly. I just had to add support for clear text traffic on the Xcode side but that is a property some people would need, NSAppTransportSecurity.

Can we get this merged with the main repo ASAP.

ryanheise:
Hi @YaarPatandarAA , I would certainly like to merge it, but the issue is maintainability. I do plan to continue adding features, and adding identical changes in two places doesn't sound appealing going forward.

I originally put the issue on hold because @stuartmorgan (above) and other members of the Flutter team have been discussing best practices for doing precisely what we want to do here, but assuming those discussions may take a while to resolve, I'd be happy to just make a decision now.

Just as food for thought, does anyone know the chances of running into a feature where it would need to be implemented differently on iOS vs MacOS? If there is a chance, then maybe copying the directory "is" the right way to go.

YaarPatandarAA:
@ryanheise Thanks for the update on the matter.

As for the question, other than Control Center related features for music playback. As far as I know there shouldn't be any feature related to audio playback that is different than iOS and MacOS. Recently Apple has been doing a lot work on the background to make sure iOS apps can run on MacOS with little to no changes.

I have been implementing and using this pull request all day today as I have been developing and I have been following the Documentation from master and have been using it as I would a iOS app and have not had any problems.

ryanheise:
@YaarPatandarAA , had there been any differences, it may have been easier to justify actually :-) But still, I'm convinced we can merge this now/soonish. If any contributor would later like to update the MacOS implementation as per the recommendations that come out of https://github.com/flutter/flutter/issues/52285 I would be happy to accept a PR.

So for now, the plan is to merge this PR. I'll need a day to find the time to do it, and collect everything else that will go into the next release.

stuartmorgan:
The recommendation from the issue linked above is to do symlinking to share identical code. There's a PR to use it in one flutter/plugins plugin as as example.

ryanheise:
Thanks!

To everyone interested in this feature, I have implemented the symlink approach following the example @stuartmorgan linked to and also using @sachaarbonel 's PR as a guide (thanks @sachaarbonel !)

I will close this PR although please let me know whether or not the committed symlink solution works for you.

YaarPatandarAA:
Was able to test from master, working fine for me. Tested the same checked off sections from iOS and they work for Mac also. Wasn't able to figure out the file path one, how do I use that? I know with Apple they are very restrictive on how apps can access directories on the device.

Otherwise I believe the readme can be updated to include a MacOS column and the same checked off boxes as iOS. 

Additionally one change I would recommend is to adjust the @avaiable() to include Mac also.
`@available(macOS 10.12, iOS 10.0, *)`

Thank you for including MacOS support.

ryanheise:
Thanks for the suggestions, @YaarPatandarAA , I have updated the documentation and available checks.

ryanheise:
Hi @sachaarbonel , you might like to know about the changes in the upcoming release detailed on #131 which add or improve various features to the web implementation, although they haven't been fully tested. I just wanted to let you know ahead of the next release in case you notice anything that breaks.

ryanheise:
Edit - sorry, I meant "macOS" implementation.

PR #78, opened 2020-04-21
Add Icy metadata support on Android

Jei:
This PR adds support to metadata retrieval from streams using the Icy specification (typically Shoutcast streams). I've tried to follow the structure of the data provided by ExoPlayer by separating it into two different classes: `IcyInfo` and `IcyHeaders`. The former contains the title and the URL provided by the stream (if provided at all). The latter contains the headers returned by the HTTP connection. They're both encapsulated into an `IcyMetadata` object, provided by the `icyMetadata` attribute of `AudioPlaybackEvent`. You can access it from `playbackEventStream`, `fullPlaybackStateStream` and `icyMetadataStream`.

- add `IcyMetadata`, `IcyInfo`, and `IcyHeaders` classes;
- add `AudioPlaybackEvent.icyMetadata` parameter;
- add `AudioPlayer.icyMetadataStream` and `AudioPlayer.icyMetadata`;
- update ExoPlayer to version 2.11.4.


ryanheise:
Nice work! It looks like it may crash on iOS and macOS which don't broadcast the additional event parameters, but I'll merge and I'll add one more commit onto this from my Mac(InCloud).

odrevet:
This is great ! I'd love to see a practical example with audio_service to set android notification and refresh it when the song change

ryanheise:
If someone feels inclined to write an example, please consider adding it to the cookbook in the wiki :-)

PR #76, opened 2020-04-19
Exoplayer exceptions

terryBaz:
- pump Expo player exceptions to flutter (android) https://github.com/ryanheise/just_audio/issues/31
- fix `.gitignore`
- update `Readme`
- code reformat `AudioPlayer.java`

PR #80, opened 2020-04-22
Forward ExoPlaybackException exceptions to the Flutter plugin

Jei:
Following up on #76, I've modified the Android code to forward any playback exception thrown by ExoPlayer to the Flutter plugin. This allows reacting to exceptions thrown while playing audio (connection errors, unexpected exceptions...).
In my current solution, the error events are only passed to `playbackEventStream`, while all the other streams ignore them.

### DIFF
--- a/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -349,6 +349,10 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Met
 			prepareResult.error(errorCode, errorMsg, null);
 			prepareResult = null;
 		}
+
+		if (eventSink != null) {
+			eventSink.error(errorCode, errorMsg, null);
+		}
 	}
 
 	private void transition(final PlaybackState newState) {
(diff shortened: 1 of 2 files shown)


stonega:
My own solution is adding a separate error string stream in AudioPlaybackEvent
```
final _playbackErrorSubject = BehaviorSubject<String>();

 _eventChannelStream = EventChannel('com.ryanheise.just_audio.events.$_id')
        .receiveBroadcastStream()
      .map((data) => _audioPlaybackEvent = AudioPlaybackEvent(
              state: AudioPlaybackState.values[data[0]],
              buffering: data[1],
              updatePosition: Duration(milliseconds: data[2]),
              updateTime: Duration(milliseconds: data[3]),
              bufferedPosition: Duration(milliseconds: data[4]),
              **playbackError: data[5],**
              speed: _speed,
              duration: _duration,
            ));
```
But would be confilict with other platform. I am only using android, so it just work.

Jei:
I've changed `playbackStateStream`, `bufferingStream`, `bufferedPositionStream` and `icyMetadataStream` to ignore oncoming errors from `playbackEventStream`. It's the only way I could think of to avoid having to write an error handler for each stream.
@stonega that's a possible solution too, but I'd prefer not to break compatibility with previous versions or with iOS. Another solution would be to add `AudioPlaybackState.error` as a possible state of `AudioPlaybackEvent`, but it would be a pretty big change to the whole plugin.

ryanheise:
Thanks, @Jei . This seems like a reasonable approach, although the best approach is often clearer in hindsight, so let's go with it and see how it goes.

Issue #84, opened 2020-04-26
Duration is not updated when playing a source which have dynamic duration (DASH/HLS)


lkho:
**Describe the bug**
When working with DASH/HLS streams, the duration can be unknown at the beginning, and being updated many times during playback. The duration returned from the setUrl() initially is not enough. Moreover, the player position is calculated with regards to the duration, thus when the duration is updated to a larger value after some time during playback, the position reporting is capped at a maximum of the initial duration.

**To Reproduce**
1. I don't have a public DASH/HLS url, but any http server which serves the output directory of a ffmpeg command e.g. `ffmpeg -i XXX -f dash -hls_playlist 1 -seg_duration 20 master.mpd` will work.  
(the manifest need to be 'live', which means it will update periodically.)
2. load the DASH/HLS url
3. the initial duration will be reported as the current length of the manifest, up to the point ffmpeg has processed.  
(it is the first segment length if you start playing immediately after you start ffmpeg)  
(you can add `-re` flag before `-i` to slow down transcoding)
4. use the getPositionStream() to track the position of the player
5. when the player position reaches that duration, the value will stay there and stops updating, even though it is still playing.
6. currently the lib does not have method to re-fetch the duration.

**Expected behavior**
the duration should be emitted back to to the UI if it is changed, so the player position can be calculated correctly.

**environment**
 - OS: android 8.0
 - Flutter SDK: `v1.12.13+hotfix.9`
 - just_audio: `0.1.7`

**related issues**
#41 the buffer value will also stop updating, after the buffer reaches the initial duration. (but it is still buffering more content, since the manifest duration has been extended.)

https://github.com/ryanheise/just_audio/blob/fdea4398e7356412e55e5b3896439c0642df0e74/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java#L84

#68 describes iOS, but on android when the duration is unknown, it also returns negative (`-9223372036854775807L` which is `C.TIME_UNSET` in exoplayer)

lkho:
I have a fork hot-fixing this issue, but is android only. See whether it can help.

https://github.com/lkho/just_audio/tree/pr/android-duration

ryanheise:
Hi @lkho , thanks for that. Certainly clipping the position to the duration should not be done when the final duration is unknown, and your solution conveniently seems to handle the iOS case as well (which was reported for plain mp3s with an unknown duration).

[...]

Issue #68, opened 2020-04-06
Negative duration returned for streaming mp3 on iOS device


suragch:
I'm streaming an mp3 file from my server. Using just_audio works as expected on an Android device, but when I try to get the audio duration on an iOS device (simulator), a negative value is returned.

This does not happen when using the [example](https://pub.dev/packages/just_audio#-example-tab-) audio file `"https://s3.amazonaws.com/scifri-episodes/scifri20181123-episode.mp3"`, so I'm not sure if this is a bug in the plugin or a bug in how my server is returning the headers. 

If it is a bug in my server setup, what are the required headers that iOS needs to obtain the mp3 audio duration?

ryanheise:
Could you provide a link to an example mp3 that illustrates the problem?

suragch:
Here is an example:

```
_totalDuration = await _player.setUrl('https://yorool.com/api/v1/audio/book/example.mp3');
print('total duration: $_totalDuration');
```

This gives the following duration:

```
total duration: -596:31:23.648000
```

The audio still plays but I can't manage the seek bar since I don't know the total duration.

rohansohonee:
@suragch

Please use a real iOS device for testing. Simulators/Emulators often don't work as expected.
I am just sharing my experience.

ryanheise:
@suragch , another thing to confirm is, when you say you are "streaming" the mp3, how are you streaming it? Do you mean you're not just offering the mp3 at a URL, but you're using some sort of streaming protocol? If so, which one?

suragch:
@rohansohonee Good point. The simulator can behave differently than a real device, though it is interesting that the android emulator works while the iOS simulator doesn't. I tested it from Xcode on an iPhone 4s (the only iOS device I have) and got a cryptic error message from Xcode:

```
2020-04-07 13:02:34.302 Runner[201:3943] time discontinuity detected: -656
```

Not even the audio played this time.

suragch:
@ryanheise I'm a newcomer to streaming. I assumed that streaming just meant setting the URL and then calling `player.play()` rather than downloading first before calling play. I'm not using any sort of streaming protocol client side other than what just_audio provides.

simdev0:
I'm also having this issue. I use .setUrl(url).then((duration) { /* set duration here */ } and found that it comes up with a negative value.

ryanheise:
> Here is an example:
> 
> ```
[...]

PR #94, opened 2020-05-21
Fix infinity duration error on web platform

khairul169:
This will fix error while trying to load url from a backend that stream audio with unknown duration on web platform.

```
Error: Unsupported operation: Infinity
    at Object.createErrorWithStack (http://localhost:65222/dart_sdk.js:4786:12)
    at Object._rethrow (http://localhost:65222/dart_sdk.js:35900:16)
    at async._AsyncCallbackEntry.new.callback (http://localhost:65222/dart_sdk.js:35896:13)
    at Object._microtaskLoop (http://localhost:65222/dart_sdk.js:35756:13)
    at _startMicrotaskLoop (http://localhost:65222/dart_sdk.js:35762:13)
```

### DIFF
--- a/lib/just_audio_web.dart
+++ b/lib/just_audio_web.dart
@@ -154,7 +154,8 @@ class Html5AudioPlayer extends JustAudioPlayer {
     _audioElement.src = url;
     _audioElement.preload = 'auto';
     _audioElement.load();
-    final duration = await _durationCompleter.future;
+    var duration = await _durationCompleter.future;
+    duration = duration != double.infinity ? duration : 0;
     transition(AudioPlaybackState.stopped);
     return (duration * 1000).toInt();
   }


kitek:
Yeah, I have got the same issue with online radio streaming sources but I was wondering if this is correct return duration = 0 in that cases? 

How this affect all things related to seeking & progress sliders?
For me personally returned value, in that case, should be different than 0, because 0 says no duration, in my opinion, better is -1 or the best is to redesign this into some type with available values like infinity, finite with duration x.

ryanheise:
Do you have a link that I can use for testing?

kitek:
Sure, use it with the library example: 
`https://uk7.internet-radio.com/proxy/danceradioukchatbox?mp=/stream;`

ryanheise:
Related to #68 and #84 .

ryanheise:
What was discussed in the other related issues above was to make the duration return null if it is unknown. Also in some streams the duration can be reported and updated over time, and it would be nice to reflect that in `just_audio`. But at the very least, I think it makes sense for duration to be either null or negative if unknown.

@lkho addressed this for Android in his fork (https://github.com/lkho/just_audio/commits/pr/android-duration) which might be informative for the various considerations here. Still not sure on the best way to proceed, but I'm leaning toward using null when the duration is unknown, and then modifying the dart code to not clip the position to the duration.

How this affects seeking, etc. is another issue, and may be related to #96 .

lkho:
> Still not sure on the best way to proceed, but I'm leaning toward using null when the duration is unknown, and then modifying the dart code to not clip the position to the duration.

My suggestion is not to break existing code, at least not crashing. Also, if `null` is a valid value, then `AudioPlaybackEvent.copyWith()` may break.

ryanheise:
> My suggestion is not to break existing code, at least not crashing. Also, if `null` is a valid value, then `AudioPlaybackEvent.copyWith()` may break.

Thanks for that, I agree we don't want to do anything radical here.

Although this could be one of the advantages of using `null` actually, because it will only break this specific use case (which is the one that needed to be fixed anyway), and any new null exceptions that get introduced during development will be informative in pointing out which parts of the code need to be checked to make this use case work correctly :-)

I'd be happy with something like this:

1. the platform side sends a null duration when the duration is unknown
2. the platform broadcasts state changes which include the duration when the duration changes (not sure how this works on web, and can be considered later)
3. the dart side does the null checking and avoids position clipping when duration unknown

By the way, the reason for clipping the position to the duration originally was because often the reported position and/or duration is wrong. Probably it is more often that the reported position is wrong, especially after a seek. And having a position greater than the duration breaks a constraint that you'd expect to have when rendering a seek bar. But I'm not really sure how a seek bar should be rendered in the case of an unknown duration anyway.

lib/just_audio_web.dart
```diff
@@ -154,7 +154,8 @@ class Html5AudioPlayer extends JustAudioPlayer {
     _audioElement.src = url;
     _audioElement.preload = 'auto';
     _audioElement.load();
-    final duration = await _durationCompleter.future;
+    var duration = await _durationCompleter.future;
+    duration = duration != double.infinity ? duration : 0;
```
ryanheise, line 158: When the duration is unknown, return -1 to have it correctly handled on the Flutter side.

ryanheise:
I've updated the duration code in the latest commit in line with how it works on iOS, so I'll close this PR.

PR #104, opened 2020-06-03
support dynamic duration

lkho:
Suggested solution for https://github.com/ryanheise/just_audio/issues/84


ryanheise:
Awesome! I appreciate the guards you added which should help to avoid a crash on web, iOS and macOS which don't yet pass the duration in the events. This will also be a good basis for the other platforms to follow suit.

Issue #105, opened 2020-06-03
Set User-Agent customization 


getmmg:
**Is your feature request related to a problem? Please describe.**

Currently the user-agent is set to just_audio and version, I have a requirement that it needs a custom header. 

**Describe the solution you'd like**
Set the UserAgent in AudioPlayer Constructor or an optional argument in setUrl

naguchennai:
This is only audio service code.. If you want change user agent or customization  you must ask [Just_audio ](https://pub.dev/packages/just_audio#-readme-tab-).. They only playing audio ... thank you

naguchennai:
Yes i also want it thank you

ryanheise:
I have implemented a proxy-based solution in the latest commit. There are still unresolved issues with the the `position<=duration` constraint being violated, although this is regarding a separate PR and I'm working on a fix for that.

Note that the proxy-based solution I implemented does not yet recursively apply headers to items in playlist formats like `m3u8` but it should work for simple URLs like `.mp3`. The solution should also work for Android, iOS, macOS, but not web (a separate solution will be needed for this platform).

ryanheise:
The unrelated issue with the duration is now resolved.

Let me know if the headers parameter works for you, or whether you need this to work for playlist formats like `m3u8`.

getmmg:
Hi I am setting Header like below 

`Map<String,String> header = {'user-agent' : 'xxx'}`

In the request I am seeing the header as `user-agent: Dart/2.8 (dart:io), xxx`, Is it possible to remove the extra and keep only the xxx as user-agent.

ryanheise:
Ah, interesting. Yes I can take a look at that tomorrow.

ryanheise:
This should be fixed in the latest commit.

ryanheise:
This is now available in the 0.2.0 release.

getmmg:
Hi @ryanheise 

I have noticed couple of issues with 0.2.0 

Getting below exception while trying to play some url's 

```
E/flutter (15845): [ERROR:flutter/lib/ui/ui_dart_state.cc(157)] Unhandled Exception: HttpException: Invalid response line, uri = http://103.16.47.70:9555/;stream.mp3
E/flutter (15845): #0      _HttpParser._doParse (dart:_http/http_parser.dart:459:15)
E/flutter (15845): #1      _HttpParser._parse (dart:_http/http_parser.dart:328:7)
[...]

Issue #99, opened 2020-05-28
Adding Headers to setUrl


Jeremiah104:
Hi I am new to flutter and I am trying to stream an audio file from amazon s3, this package seems to be able to do everything I need when I use a public link but I need to access a link that uses headers for authentication and I can't find where to add headers.

I'm not sure if there is a simple solution to this that I am unaware of, or if it is a functionality that would need to be added.

shipofsea:
Hello, 
can you let me know the audio file url for test?
thanks

shipofsea:
I have read your post on upwork.
so I send this message to you here

Jeremiah104:
Hi thanks for responding,
The url is https://moc-app-test.s3.amazonaws.com/audios/test.mp3. 

The audio file URL is inaccessible because you need the authentication headers. This is the goal of my question here - I need to be able to add S3 authentication headers to this package so that I can access and stream a non public file. 
(I am using https://github.com/diagnosia/flutter_aws_s3_client to generate the header info)

shipofsea:
Hi, can you provide AWS S3 key for me to test?
I tested stream play with a mp3 from anther server.

Jeremiah104:
Yeah I just don't want to post that publicly, what is your Upwork name/id? We can talk there and I can get you the keys.

shipofsea:
ok, then let's discuss via skype
my skype id is live:fc722aec9c3a7bb4
please let you send message here.

ryanheise:
Hi @Jeremiah104 and @shipofsea 

Is there a reason why the issue is now closed?

This is a useful feature of benefit to the larger user base and I was going to work on it, but I wasn't sure if @shipofsea had started working on it and didn't want to interfere with that particularly if it was being done through upwork or some other arrangement.

Note that there is now a pull request that implements this feature by @raveesh-me here: #107 , which provides guidance on how such a feature could be added.

Since @Jeremiah104 closed this issue without a solution being added to the repo, please track the new open issue #105 (which is essentially the same) if you are still interested in this feature, and I will do an implementation based on @raveesh-me 's suggestions.

raveesh-me:
Hi @ryanheise 
This was done by Cookytech team on Upwork. @Jeremiah104 is using a fork of our fork of your project, and we decided to send a pull request to share the results with you.

Jeremiah104:
Hi @ryanheise,
[...]

PR #107, opened 2020-06-07
[Changes iOS implementation] Add HTTPS 

raveesh-me:
This is work we did in response to  #99.
#105 seems similar

This solution is not necessarily compatible. Especially the iOS implementation, but we hope this helps you add a new feature.


ryanheise:
Thanks, @raveesh-me . This is definitely helpful, particularly as I was not familiar with the relevant iOS API.

For reasons unknown, #99 was closed, but I will take #105 as the relevant issue since it is still open.

raveesh-me:
> For reasons unknown, #99 was closed, but I will take #105 as the relevant issue since it is still open.
#99 was closed because we fixed it for them.
#105 can still use our fork for the meanwhile

>  I was not familiar with the relevant iOS API.
Our implementation uses a private call which is not documented in the iOS development docs. So I guess for this one even Apple is not aware of the relevant APIs on the iOS side. :-p

ryanheise:
> #99 was closed because we fixed it for them.

Ah, thanks for explaining that. So "Biswajit Behera" in the git logs owns the contributions, or is it cookytech? (I ask in relation to section D6 of the GitHub terms of service.)

> > I was not familiar with the relevant iOS API.
> 
> Our implementation uses a private call which is not documented in the iOS development docs. So I guess for this one even Apple is not aware of the relevant APIs on the iOS side. :-p

Oh, that does complicate things, since the Apple developer license agreement may prevent anyone from using this plugin if it uses undocumented APIs. I may need to implement an alternative such as creating a local HTTP proxy on device. This may make my previous question mute.

raveesh-me:
>Ah, thanks for explaining that. So "Biswajit Behera" in the git logs owns the contributions, or is it cookytech? (I ask in relation to section D6 of the GitHub terms of service.)

"Cookytech" will be the owner.

> Oh, that does complicate things, since the Apple developer license agreement may prevent anyone from using this plugin if it uses undocumented APIs. I may need to implement an alternative such as creating a local HTTP proxy on device. This may make my previous question mute.

Oh!

ryanheise:
I've implemented a partial proxy-based solution in the latest commit, so I'll close this.

Issue #96, opened 2020-05-25
Go back to live audio after pause or stop


diego-lipinski-de-castro:
**Describe the solution you'd like**
When streaming, I would like a way to go back to the currently live audio after the user has paused or stopped the player.

Currently, If i stop or pause the player, and play again, it will play the audio as a playback, not what is live at the moment, and I cant find a way to stop that.

Im not sure If Im clear on the problem, Im not sure how I can explain a better way

lkho:
Feature suggestion: add a way to seek to the end of live streams

Android:
`Player.seekToDefaultPosition()` or `Player.seekTo(C.TIME_UNSET)`
iOS:
https://stackoverflow.com/a/13892727
`[player seekToTime: CMTimeMakeWithSeconds(MAXFLOAT, NSEC_PER_SEC)];`

ryanheise:
Hi @diego-lipinski-de-castro @lkho 's fix is now available in the latest 0.2.0 release. To use it, call `player.seek(null)` to seek to the end of the live stream. Let me know how it goes.

diego-lipinski-de-castro:
Worked on Android, on iOS (13.5) it does nothing tho

ryanheise:
Hi @diego-lipinski-de-castro 

On line 77 of `ios/Classes/AudioPlayer.m` would you be able to try changing `kCMTimePositiveInfinity` to `CMTimeMakeWithSeconds(MAXFLOAT, NSEC_PER_SEC)` and let me know if it works in your case?

Another alternative to try is to replace lines 77-78 with:

```objc
if (args[0] == [NSNull null]) {
    NSValue *value = player.currentItem.seekableTimeRanges.lastObject;
    if (value) {
        CMTimeRange seekableRange = [value CMTimeRangeValue];
        CMTime latestTime = CMTimeRangeGetEnd(seekableRange);
        [self seek:latestTime result:result];
    }
} else {
    [self seek:CMTimeMake([args[0] intValue], 1000)];
}
```

CpHeat:
Just to let you know seek(null) is not working on iOS. For now I have to do a stop() then play() to get to the actual live but it has a bad feeling since it puts the notification on offline mode for a sec (on iOS only). I also tried with a fantaisist seek(Duration(hours: 99)) but it didn't do nothing on iOS either, it just played back to where it paused like seek(null) does. I'm experimenting on iOS 14
That could be a very nice feature to implement if you could

ryanheise:
[...]

PR #108, opened 2020-06-07
support seeking to end of live streams

lkho:
suggested edit for #96

Not tested in iOS. Feel free to edit.


ryanheise:
Thanks for this. I can have a go at the iOS implementation after finishing off a different issue.

PR #109, opened 2020-06-07
Android: Support Android Embedding V2

ened:
Migrate to the new Android embedding which helps to make the lifecycle & memory cleanup a lot more explicit & efficient.
This also resolved a memory leak as discovered by LeakCanary.


android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -500,7 +505,11 @@ public void seek(final long position, final Result result) {
...
 
 	public void dispose() {
-		player.release();
+    if (player != null) {
+      player.release();
+      player = null;
+		}
+
 		buffering = false;
 		transition(PlaybackState.none);
```
ryanheise, line 514: These should go inside the `if` since the `dispose` code should only be executed once.

android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -500,7 +505,11 @@ public void seek(final long position, final Result result) {
 	}
 
 	public void dispose() {
-		player.release();
+    if (player != null) {
```
ryanheise, line 508: This should be tabbed indentation to be consistent with the rest of the file.

android/src/main/java/com/ryanheise/just_audio/JustAudioPlugin.java
```diff
@@ -1,41 +1,56 @@
...
-  /** Plugin registration. */
-  public static void registerWith(Registrar registrar) {
-    final MethodChannel channel = new MethodChannel(registrar.messenger(), "com.ryanheise.just_audio.methods");
-    channel.setMethodCallHandler(new JustAudioPlugin(registrar));
+/**
+ * JustAudioPlugin
+ */
+public class JustAudioPlugin implements FlutterPlugin {
+
+  private MethodChannel channel;
```
ryanheise, line 15: Use tabbed indentation here, too.

ened:
@ryanheise addressed both your comments. What AS config are you using to ensure consistent source code style? I'm used to the google-java-format config that the Flutter project uses.

android/src/main/java/com/ryanheise/just_audio/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+import io.flutter.plugin.common.BinaryMessenger;
+import io.flutter.plugin.common.MethodCall;
+import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
+import io.flutter.plugin.common.MethodChannel.Result;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MethodCallHandlerImpl implements MethodCallHandler {
+
+  private final Context applicationContext;
```
ryanheise, line 14: Tabs

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.1.11
```
ryanheise, line 1: Unfortunately I can't release this yet. I have unstable features on master that are in development. The next release will be a breaking change and will bump to 0.2 but leave this out for now as I will update the change log when the next release is ready to roll out.

ened: @ryanheise ok no problem. Would you like me to remove the change log entry & version bump from this PR so that it's still mergeable?

android/src/main/java/com/ryanheise/just_audio/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+				break;
+		}
+	}
+
+	void dispose() {
+		for (AudioPlayer player : players) {
+			player.dispose();
+		}
+
+		players.clear();
```
ryanheise, line 48: I understand the benefits of maintaining this list of players, although it appears that this list grows ever larger until the flutter engine is destroyed. Ideally player instances would be removed from this list as soon as an app calls `player.dispose()` from the Flutter side. It is perhaps not as devastating as it may otherwise be to leave these players in memory since just its just the shell that remains in memory, while the heavyweight ExoPlayer kernel is at least disposed of. It is tempting to handle "dispose" on the main channel here where you would have access to the players list but that would require a change to the other platform implementations + testing. A more isolated way is to pass a reference to `this` into the new player which it can later call back on during `dispose` to do this cleanup.

If it sounds not worth it, don't worry, you can leave it as is and I might look into this again later.

ryanheise:
Hi @ened , thanks and great work on this PR. Ensuring player instances are disposed on the platform side is definitely a good direction to go in.

Regarding the coding style, I use tabs for indentation in Java, and I don't indent case branches in switch statements (this is a rather old school preference from SUN's original Java coding style guidelines). I think I've got all IDE clutter in the .gitignore file. I assume AS means Android Studio? If so, I assume the TABs option should be easy to find, and you want to deselect the "Indent case branches" option.

I may consider caving to the pressure of the IDEs' recommended style at some point in the future but I'd rather do such a change in a separate PR.

pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: just_audio
 description: Flutter plugin to play audio from streams, files, assets and DASH/HLS streams. Works with audio_service to play audio in the background.
-version: 0.1.10
+version: 0.1.11
```
ryanheise, line 3: Leave this unchanged for now as per my other comment.

CHANGELOG.md
ryanheise: Yes please leave this out as I'd rather manage the change log when I do the next release. Thanks.

android/src/main/java/com/ryanheise/just_audio/MethodCallHandlerImpl.java
ened: How about moving the dispose method to the MethodCallHandlerImpl, where it can manage the removal of the list item and calls the dispose method on the AudioPlayer object from the outside?

ryanheise: That's what I meant by "It is tempting to handle dispose ... here" in my comment above. But this method handler class (which should probably be named in a less anonymous way - `MainMethodCallHandler`?) is for the main method channel and moving this method from one channel to another will impact the other platform implementations (iOS and web). I'd prefer to avoid making a change with such impact unless it's fully thought through, so I'd prefer for now something that only affects the Android implementation.

ryanheise: Oh, i see you already did that. But you see, this will break the iOS and web implementations - isolating this to just the Android implementation will be better for now, and the bigger picture can be looked at afterwards.

ened: @ryanheise ok, let me sort it and update the PR again.

ened: @ryanheise updated once again.

ryanheise: Great, assuming that's good I think it's ready to merge, but it's 3:50am so I'm going to sleep and will take a fresh look in the morning :-)

android/src/main/java/com/ryanheise/just_audio/MethodCallHandlerImpl.java
```diff
@@ -0,0 +1,52 @@
...
+			result.notImplemented();
+			break;
+		}
+	}
+
+	void dispose() {
+		final int size = players.size();
+		for (int i = 0; i < size; i++) {
+			final AudioPlayer player = players.valueAt(i);
+			player.dispose();
```
ryanheise, line 47: This will cause the size of the array to become shorter while iterating, and may cause every second player to be skipped and not get disposed. You might want to iterate over a clone of the array to iterate over a stable collection.  If `dispose` triggers the runnable `players.remove(id)` to be called, then also the `players.clear()` line below is redundant.

ened: Ok I forgot about that. Will update..

ened:
@ryanheise updated the code once again

lib/just_audio.dart
```diff
@@ -451,7 +451,7 @@ class AudioPlayer {
   /// * [AudioPlaybackState.none]
   /// * [AudioPlaybackState.connecting]
   Future<void> dispose() async {
-    await _invokeMethod('dispose');
+    await _invokeMethod('dispose', [_id]);
```
ryanheise, line 454: This change is not necessary (actually the parameter is unused).

android/src/main/java/com/ryanheise/just_audio/MethodCallHandlerImpl.java
ryanheise: The `players.clear()` line below is still redundant, though.

ryanheise: Also, `players.values()` does not make a stable clone of the map. It's actually backed by the map, so any changes to the map will appear in `values`.

android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -270,14 +270,15 @@ public void onMethodCall(final MethodCall call, final Result result) {
...
+					position2 = (Integer) position;
 				} else {
-					position2 = (Long)position;
+					position2 = (Long) position;
 				}
 				seek(position2 == -2 ? C.TIME_UNSET : position2, result);
 				break;
 			case "dispose":
 				dispose();
+				onDispose.run();
```
ryanheise, line 281: This call to `onDispose.run()` is redundant since the `dispose()` line just above already calls this.

ryanheise:
Hi @ened I left some comments but I've decided to merge. The changes I proposed are changes I will be able to make myself afterwards. Thank you greatly for the PR!

ened:
OK, great. Happy the plugin is improved a bit. :)

ryanheise:
I've committed some minor changes as per my comments, although not fully tested. Let me know if things break for you :-)

PR #142, opened 2020-08-09
Fixed concatenating audio source methods

nuc134r:
Tried to implement playlist reordering logic using ConcatenatingAudioSource and a bunch of `(Untested)`-prefixed methods failed.

First they failed due to expected String key in `mediaSources` map. I removed the cast completely as `Map<,>.get()` expects `Object`. Not sure if it should be this way, maybe we better cast it to `String`.

Then ExoPlayer's ConcatenatingMediaSource methods expected both `Handler` and `Runnable` not null or both null in a runtime assertion. Passed `handler` from local scope.

This fixed reordering and clearing ConcatenatingAudioSource issues for me but I didn't test other methods. And I'm not a Java/Android dev so I could do something stupid while fixing my issues, sorry if it's so.

### DIFF
--- a/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -319,34 +319,34 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Met
 				break;
 			case "concatenating.add":
 				concatenating(args.get(0))
-						.addMediaSource(getAudioSource(args.get(1)), null, () -> result.success(null));
+						.addMediaSource(getAudioSource(args.get(1)), handler, () -> result.success(null));
 				break;
 			case "concatenating.insert":
 				concatenating(args.get(0))
-						.addMediaSource((Integer)args.get(1), getAudioSource(args.get(2)), null, () -> result.success(null));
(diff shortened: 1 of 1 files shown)


ryanheise:
Thanks! The `handler` parameter that you've added is definitely correct. And you're also right that the cast was supposed to be to `String`. If you can make that change, I'll be happy to merge this.

nuc134r:
Done. You're the best.

ryanheise:
Thank you, merged!

ryanheise:
Incidentally, are you also using this feature on iOS, and did it work for you? The iOS implementation is also untested, and I would be very surprised if it actually works out of the box without ever having been tested. These features don't exist natively on iOS, so there is greater complexity in the code to emulate them.

nuc134r:
No, sorry. Currently owning and targeting only Android devices.

ryanheise:
I just did some of my own testing on Android and your PR works nicely. However, I did notice that there was no way for an app to listen to changes to the contents of a playlist, so I added two new streams: `sequenceStream` which broadcasts the current list of `IndexedAudioSource`s, and `sequenceState` stream which broadcasts the pair of the current sequence and current index within that sequence. A guarantee is provided that the components of the `sequenceState` are updated consistently, so if the current index is right on the last item in the playlist, and then you delete it, you won't temporarily have the index pointing out of bounds, they will both update at the same time.

PR #170, opened 2020-08-31
Relative seeking, position and duration for iOS

snaeji:
These changes add support for HLS event playlist type.

I was not sure if I could access the seekableTimeRanges in the indexedAudioSource, so we get it from the player. If you have any other preferred solution please let me know.

I tested this implementation with regular audio files and everything works as intended.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.1
+
+* Fix setSpeed bug on iOS.
+
 ## 0.4.0
 
 * Handles audio focus/interruptions via audio_session
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: just_audio
 description: Flutter plugin to play audio from streams, files, assets, DASH/HLS streams and playlists. Works with audio_service to play audio in the background.
-version: 0.4.0
+version: 0.4.1
 homepage: https://github.com/ryanheise/just_audio
 
 environment:


ryanheise:
Nice work!

> The only thing I wasn't sure about is if it's okay to access the currentItem straight from the player. It works but i guess it would be cleaner if we could get it from the indexedAudioSources.

Yes I agree something along those lines would be cleaner. Not necessarily accessing the `currentItem` property but accessing the indexed audio source's `playerItem` property which stores the player item associated with that source (which is the only player item that the source should be concerned about).

snaeji:
> Yes I agree something along those lines would be cleaner. Not necessarily accessing the `currentItem` property but accessing the indexed audio source's `playerItem` property which stores the player item associated with that source (which is the only player item that the source should be concerned about).

I just added a commit that uses the playerItem. Tested and works great 🎉 

Is there anything else that you would like to change or consider?

darwin/Classes/AudioPlayer.m
```diff
@@ -335,7 +335,9 @@ - (int)getCurrentPosition {
     } else if (CMTIME_IS_VALID(_seekPos)) {
         return (int)(1000 * CMTimeGetSeconds(_seekPos));
     } else if (_indexedAudioSources && _indexedAudioSources.count > 0) {
-        int ms = (int)(1000 * CMTimeGetSeconds(_indexedAudioSources[_index].position));
+        CMTimeRange seekableRange = [_indexedAudioSources[_index].playerItem.seekableTimeRanges.lastObject CMTimeRangeValue];
+        CGFloat seekableStart = CMTimeGetSeconds(seekableRange.start);
+        int ms = (int)(1000 * CMTimeGetSeconds(_indexedAudioSources[_index].position)) - (seekableStart * 1000);
```
ryanheise, line 340: I think this needs to be encapsulated inside of the relevant indexed audio source. In other words, leave this line as per the original:

```dart
int ms = (int)(1000 * CMTimeGetSeconds(_indexedAudioSources[_index].position))
```

but then allow the implementation of the `position` property to figure out the correct position, since even the seekable time ranges will be offset for a `ClippingAudioSource`. All you would need to do is push this code into `UriAudioSource`. If you look at the implementation of `ClippingAudioSource` you can see that it will automatically benefit from this code, and then correctly offset the position.

The same applies to `duration` and `seek` which are also encapsulated within the relevant indexed audio source because the behaviour of each of these is dependent on what type of indexed audio source it is.

snaeji:
Sorry about closing the PR, i messed up my branch but a new PR is on it's way!

ryanheise:
Phew!! Look forward to it :-)

snaeji:
Here is the new PR :)
https://github.com/ryanheise/just_audio/pull/181/commits/1d17025d5dca1035a2fea763c9832a076c8f24b7

PR #176, opened 2020-09-03
bug fix for audio seek when playback speed > 1.0

subhash279:
When seeking audio at playback rate > 1.0, after the seek, the playback rate was defaulting to 1.0
This should fix it and also make the seek slightly faster.
Please take a look!

### DIFF
--- a/darwin/Classes/AudioPlayer.m
+++ b/darwin/Classes/AudioPlayer.m
@@ -1109,6 +1109,7 @@
                     [_player playImmediatelyAtRate:_speed];
                 } else {
                     [_player play];
+                    _player.rate = _speed;
                 }
             } else {
                 // If not playing, there is no reliable way to detect
--- a/darwin/Classes/AudioPlayer.m
+++ b/darwin/Classes/AudioPlayer.m
@@ -1105,7 +1105,11 @@
                 // If playing, buffering will be detected either by:
                 // 1. checkForDiscontinuity
                 // 2. timeControlStatus
-                [_player play];
+                if (@available(iOS 10.0, *)) {
+                    [_player playImmediatelyAtRate:_speed];
+                } else {
+                    [_player play];
+                }
             } else {
                 // If not playing, there is no reliable way to detect
                 // when buffering has completed, so we use
--- a/darwin/Classes/AudioPlayer.m
+++ b/darwin/Classes/AudioPlayer.m
@@ -1108,7 +1108,6 @@
                 if (@available(iOS 10.0, *)) {
                     [_player playImmediatelyAtRate:_speed];
                 } else {
-                    [_player play];
                     _player.rate = _speed;
                 }
             } else {


darwin/Classes/AudioPlayer.m
```diff
@@ -1105,7 +1105,11 @@ - (void)seek:(CMTime)position index:(NSNumber *)newIndex completionHandler:(void
                 // If playing, buffering will be detected either by:
                 // 1. checkForDiscontinuity
                 // 2. timeControlStatus
-                [_player play];
+                if (@available(iOS 10.0, *)) {
+                    [_player playImmediatelyAtRate:_speed];
+                } else {
+                    [_player play];
```
ryanheise, line 1111: If it fails to resume playback at the correct speed, for this else case how about we do this?

```dart
    _player.rate = _speed;
```

subhash279: yeah that's a good point. That should work.
Sry since most devices have iOS>10, i forgot that bit.

subhash279: done!

ryanheise:
Looks good. There are actually a few places throughout the code where I used `[_player play]` and perhaps these should all be replaced by similar code to your PR. But I'll merge this for now and if you notice any further problems in relation to the selected speed not being restored (e.g. after a seek) let me know.

ryanheise:
OK, just one thing before I merge, though, it is unnecessary to call `[_player play]` just before setting the rate property because setting the rate property has the effect of playing. Can you test deleting that line since you added `_player.rate = _speed`?

subhash279:
ah ok. Didn't know that. I just deleted that line.

ryanheise:
It seems good to me. I won't be able to test this myself until later, but if you have tested it already let me know and I can go ahead and merge (< iOS 10 also)

subhash279:
I tested on iOS > 10 and it works but I don't have any < iOS 10 devices/simulators to test.

ryanheise:
I'm just testing this now, and it does handle the case of seeking within a track, but it does not handle the case of seeking to a different track. As mentioned above it may be necessary to change all of the other places in the code that use `[_player play]` to `_player.rate = _speed`. I'm not sure if you want to have a crack at this or if you're happy to let me do it. I'll be away from my computer for about an hour, and I can update these other places in the code (unless you've done it already).

ryanheise:
I'm back, so I'll merge this and then add some follow-up commits.

ryanheise:
I've just updated the other spots to also use `_player.rate = _speed`. I also ended up commenting out the `playImmediatelyAtRate` case, using `_player.rate = _speed` for now, although I would like to use `playImmediatelyAtRate` as soon as I figure out a solution to an issue I noticed. The problem is that `playImmediatelyAtRate` will indeed play immediately if there is audio in the buffer, but if there is no audio in the buffer it will stall "without" notifying any observer that it is stalling. This means that I can't correctly report the buffering state while it's stalled. Maybe there is a clever way to query whether there is at least some audio in the buffer before calling it. I have tried querying `_player.currentItem.playbackBufferEmpty` to help determine the buffering state although in my experience this value wasn't 100% reliable but that would be the first thing to experiment with to implement this properly.

Anyway, for now I am using `_player.rate = _speed`, and this can be improved in future releases.

PR #186, opened 2020-09-17
Fix Airplay crash issue

snaeji:
I found a bug in my last PR.

If we already have a route to an Airplay device, and start playing. The seekableTimeRanges can be empty.
I suspect this is an AppleTV bug, but we should guard still guard against an empty array there.

This implementation only calculates a relative position if we have an object in seekableTimeRanges.

Also one minor fix where i was too enthusiastic typing semicolons.

Tested for two days 👍


ryanheise:
All looks good, thanks again!

PR #190, opened 2020-09-25
[iOS] Solving crashes when audio completes

viniciuszani:
Given the plugin architecture, `removeItemObservers` might be called twice for the same `playerItem` (when it gets completed in a queue and when the audio plugin is disposed).

This PR fixes the crash that is happening in this scenario.

### DIFF
--- a/darwin/Classes/AudioPlayer.m
+++ b/darwin/Classes/AudioPlayer.m
@@ -367,13 +367,17 @@
 }
 
 - (void)removeItemObservers:(AVPlayerItem *)playerItem {
-    [playerItem removeObserver:self forKeyPath:@"status"];
-    [playerItem removeObserver:self forKeyPath:@"playbackBufferEmpty"];
-    [playerItem removeObserver:self forKeyPath:@"playbackBufferFull"];
-    //[playerItem removeObserver:self forKeyPath:@"playbackLikelyToKeepUp"];
-    [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemDidPlayToEndTimeNotification object:playerItem];
-    [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemFailedToPlayToEndTimeNotification object:playerItem];
-    [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemPlaybackStalledNotification object:playerItem];
+    @try {
+        [playerItem removeObserver:self forKeyPath:@"status"];
+        [playerItem removeObserver:self forKeyPath:@"playbackBufferEmpty"];
+        [playerItem removeObserver:self forKeyPath:@"playbackBufferFull"];
+        //[playerItem removeObserver:self forKeyPath:@"playbackLikelyToKeepUp"];
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemDidPlayToEndTimeNotification object:playerItem];
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemFailedToPlayToEndTimeNotification object:playerItem];
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:AVPlayerItemPlaybackStalledNotification object:playerItem];
+    } @catch (id exception) {
+        NSLog(@"Tried to remove an observer for a key that did not have one");
+    }
 }
 
 - (void)addItemObservers:(AVPlayerItem *)playerItem {


ryanheise:
Thanks, @viniciuszani . Would you by any chance be able to tell me how to reproduce this crash?

viniciuszani:
Hi @ryanheise, the use case is this one (on iOS, emulator or physical device):

1. Start playing with `await setUrl();`
2. Seek the player to near completion
3. When the stream ends, the completion will trigger `removeItemObservers`, and `onStop` will do it as well.

Below is a piece of the implementation of `BackgroundAudioTask` (stripped some stuff to make it shorter).
Note that there's a chance of `onStop` being called twice in a roll (should we address this scenario with some internal vars?) if the main thread's `Stream` subscription detects the completion and calls `onStop` as well.

(but the call to onStop for a stopped stream should be a no-op from the plugin, correct?)


```dart
class ExampleBackgroundTask extends BackgroundAudioTask {
  AudioPlayer _audioPlayer = AudioPlayer(handleInterruptions: false);
  MediaItem _currentMediaItem;
  Duration _duration;
  Duration _latestPosition;

  StreamSubscription<PlayerState> _playbackStateStreamSubscription;
  StreamSubscription<PlaybackEvent> _eventSubscription;
  StreamSubscription<Duration> _positionSubscription;
  StreamSubscription<Duration> _durationSubscription;
  StreamSubscription _audioInterruptionSubscription;
  StreamSubscription _headphonesUnpluggedSubscription;

  bool get _playing => AudioServiceBackground.state.playing;
  bool _interrupted = false;

  @override
  Future<void> onStart(
    Map<String, dynamic> params
  ) async {
    _playbackStateStreamSubscription = _audioPlayer.playerStateStream
      .where((state) => state.processingState == ProcessingState.completed)
      .listen((state) {
      onStop();
    });

    _eventSubscription = _audioPlayer.playbackEventStream.listen((event) {
      _broadcastState();
    });

    _positionSubscription = _audioPlayer.positionStream.listen((position) {
      _latestPosition = position;
    });

    _durationSubscription= _audioPlayer.durationStream.listen((duration) {
      _duration = duration;
      _currentMediaItem = _currentMediaItem.copyWith(
        duration: _duration
      );
      AudioServiceBackground.setMediaItem(_currentMediaItem);
      _broadcastState();
    });

    final session = await AudioSession.instance;
    _audioInterruptionSubscription = session.interruptionEventStream.listen((event) {
      if (event.begin) {
        if (_playing) {
          onPause();
          _interrupted = true;
        }
      } else {
        switch (event.type) {
          case AudioInterruptionType.duck:
          case AudioInterruptionType.pause:
            if (!_playing && _interrupted) {
              onPlay();
            }
            break;
          case AudioInterruptionType.unknown:
            break;
        }
        _interrupted = false;
      }
    });

    _headphonesUnpluggedSubscription = session.becomingNoisyEventStream.listen((_) {
      if (_playing) {
        onPause();
      }
    });
  }

  @override
  Future<void> onClose() => onStop();

  @override
  Future<void> onTaskRemoved() => onStop();

  @override
  Future<void> onStop() async {
    try {
      await _audioPlayer.stop();
      await _audioPlayer.dispose();
      _audioInterruptionSubscription?.cancel();
      _headphonesUnpluggedSubscription?.cancel();
      _playbackStateStreamSubscription?.cancel();
      _eventSubscription?.cancel();
      _positionSubscription?.cancel();
      _durationSubscription?.cancel();

      /// It is important to wait for this state to be broadcast before we shut
      /// down the task. If we don't, the background task will be destroyed before
      /// the message gets sent to the UI.
      await _broadcastState();
    } catch (_) {}
    // mandatory to happen
    await super.onStop();
  }

  @override
  Future<void> onAddQueueItem(MediaItem mediaItem) async {
    _currentMediaItem = mediaItem;

    try {
      if (mediaItem.id.startsWith('http')) {
        await _audioPlayer.setUrl(mediaItem.id);
      } else if (mediaItem.id.startsWith('assets')) {
        await _audioPlayer.setAsset(mediaItem.id);
      } else {
        await _audioPlayer.setFilePath(mediaItem.id);
      }

      onPlay();
    } catch (e) {
      onStop();
    }
  }
}
```

ryanheise:
> Below is a piece of the implementation of `BackgroundAudioTask` (stripped some stuff to make it shorter).
> Note that there's a chance of `onStop` being called twice in a roll (should we address this scenario with some internal vars?) if the main thread's `Stream` subscription detects the completion and calls `onStop` as well.
> 
> (but the call to onStop for a stopped stream should be a no-op from the plugin, correct?)

That sounds like a good strategy for both plugins. audio_service should prevent anything bad happening from calling `onStop` twice, and just_audio should prevent anything bad happening from calling `removeItemObservers` twice.

I'll look into the just_audio issue now and see if there's some logic that needs to improve in addition to safeguarding `removeItemObservers`. Thanks for taking the time to provide the example above.

ryanheise:
> 1. Start playing with `await setUrl();`
> 2. Seek the player to near completion
> 3. When the stream ends, the completion will trigger `removeItemObservers`, and `onStop` will do it as well.

Looking at the plugin code, I can't see that the completion itself will trigger `removeItemObservers`. Rather, it will be triggered by `dispose`. The iOS implementation currently does very little checking that you are calling methods from the correct state so you currently have to ensure this yourself (i.e. don't call dispose if it is already disposed).

It's about time I do add state checking to the iOS implementation. I'll have a go at it and let you know once I've done so.

ryanheise:
I have just added some state checking and I would be curious to know if this fixes the iOS crash for you. Can you test git master? While I haven't added a try/catch yet, I would like to first fix the underlying problem, and as a last step will also consider adding an extra try/catch.

diego-lipinski-de-castro:
how can i test this on master?

diego-lipinski-de-castro:
it fixes the error for me

ryanheise:
I have also just committed some code on audio_service to ignore duplicate calls to `super.onStop`. Note that it is preferred to not call `onStop` if it is already stopped:

```dart
if (!stopped) onStop();
```

One way of checking whether it's already stopped is if you update the processing state to stopped. Then you can check it as follows:

```dart
bool get stopped => AudioServiceBackground.state.processingState == AudioProcessingState.stopped;
```

ryanheise:
I think regarding the try/catch fallback, probably a good place to add it is around `handleMethodCall` so that any unforeseen bug in the plugin gets caught and propagated to the Flutter layer rather than crashing.

ryanheise:
OK, the latest commit includes the try/catch fallback in `handleMethodCall` as mentioned above. Unless either of you notice any further issues with it, I will include this in the next release.

viniciuszani:
Hi @ryanheise, sorry the delay for the feedback.
The issue is solved with the current master version.

ryanheise:
Great! Thanks for the report. I will close this then and will include the fix in the next release.

Issue #103, opened 2020-06-01
Support Microsoft Windows ?


chuoichien:
Sorry, is this plugin support microsoft windows ?

sachaarbonel:
Some of my research notes on the subject. We could make FFI bindings to:
 "[ASIO](https://www.steinberg.net/en/company/developers.html) is an audio driver protocol by Steinberg. While it is available on multiple operating systems, it is most commonly used on Windows to work around limitations of WASAPI including access to large numbers of channels and lower-latency audio processing.". This the design decision the cross-platform Rust library [Cpal](https://github.com/RustAudio/cpal) made.

hacker1024:
I may attempt this.
@ryanheise, would using Windows code from [Musikcube](https://github.com/clangen/musikcube/tree/master/src/core) be okay? It has a BSD 3-Clause "New" or "Revised" License.

Musikcore supports gapless playback, so it looks like a good fit.

It's also written in C++, so if FFI bindings are done right, we could share the code between Windows and Linux (and even macOS, if it's better than AVPlayer).

hacker1024:
I've looked at Musikcube / Musikcore a little more - the core is completely undocumented, but there's a plugin system. There exists a documented plugin that exposes a web server to interface with the core.

It shouldn't be too hard to modify the server to use Flutter platform channels instead of HTTP, and then all the plugin documentation will apply.

https://github.com/clangen/musikcube/wiki/remote-api-documentation

If we were to uses Musikcore, that leaves two choices:
- Use FFI and use the core library directly
- Use the platform channel method I just described

ryanheise:
The Musikcore license still looks complicated to me since the individual decoders have separate license agreements (including the GPL). It's also not clear to me what the codec support is for each platform target. ExoPlayer is mentioned, for example, but ExoPlayer taps into hardware decoders that are available on device via the Android SDK, so I would probably expect to see a different set of codecs supported for each platform.

hacker1024:
That's difficult, then.
On another note, Musikcore uses the Visual Studio build system on Windows. I've been trying to build a CMake script instead (since Flutter uses CMake), but it's proving difficult.

Unfortunately, there aren't many high-level C/C++ Windows audio libraries out there...

alexmercerind:
[...]

PR #191, opened 2020-09-26
SkipSilence and volume boost options for Android

subhash279:
I implemented the above for Android. It seems more involved iOS side. So someone with more ObjectiveC experience can probably take it up in future.
Please let me know if anything else is needed.

### DIFF
--- a/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -2,7 +2,9 @@ package com.ryanheise.just_audio;
 
 import android.content.Context;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Handler;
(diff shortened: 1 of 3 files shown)


ryanheise:
Thanks! I'm currently in the process of converting this plugin into a federated plugin which I will commit first, so we will have to resolve the clash after that.

subhash279:
sure thanks!

ryanheise:
Hi @subhash279 

I've just finished the conversion to a federated plugin with a description of the changes in the following comment:

https://github.com/ryanheise/just_audio/issues/103#issuecomment-699614173

Under the new model, it is necessary to also add corresponding methods to the platform interface, and following the style of the other methods, each method in the interface would take a single message as a parameter and return a single message as a result. The web implementation just implements this interface directly while iOS and Android use the method channel version.

subhash279:
@ryanheise pls close this PR as I submitted another one which is in sync with latest plugin changes #234

PR #194, opened 2020-10-02
[FIX] - use _overrideUri in AudioSourceMessage after the assets is loaded in the tmp dir

ceppelli:
I was trying to play m4a/mp3 audio files from the assets dir on a MacOs Flutter app without success.
After some code analysis I found out the following: the method `Future<File> _loadAsset(String assetPath) async {`
copies the file from di bundle/assets dir to a tmp dir but the AudioSourceMessage sent to the native underline code doesn't contains the new tmp file location.

Regards
Luca

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -1244,7 +1244,7 @@ class ProgressiveAudioSource extends UriAudioSource {
 
   @override
   AudioSourceMessage _toMessage() => ProgressiveAudioSourceMessage(
-      id: _id, uri: uri.toString(), headers: headers);
+      id: _id, uri: (_overrideUri ?? uri).toString(), headers: headers);
 }
 
 /// An [AudioSource] representing a DASH stream. The following URI schemes are


ryanheise:
Thanks for the PR. Prior to 0.5.0, this was handled in the superclass `UriAudioSource` by:

```dart
  @override
  Map toJson() => {
        'id': _id,
        'type': _type,
        'uri': (_overrideUri ?? uri).toString(),
        'headers': headers,
      };
```

So to restore the pre 0.5.0 behaviour, the same fix would need to be applied to the `DashAudioSource` and `HlsAudioSource` subclasses. (Not that DASH or HLS are good use cases for asset URIs, but let's add it for uniformity anyway).

ceppelli:
Perfect!!! I'm glad to have been helpful.

Just one question, i'm curious: which is the reason for copying the audio file from the bundle folder to the tmp one? 

Thanks

just_audio/lib/just_audio.dart
```diff
@@ -1244,7 +1244,7 @@ class ProgressiveAudioSource extends UriAudioSource {
 
   @override
   AudioSourceMessage _toMessage() => ProgressiveAudioSourceMessage(
-      id: _id, uri: uri.toString(), headers: headers);
+      id: _id, uri: (_overrideUri ?? uri).toString(), headers: headers);
```
ryanheise, line 1247: Best to add this to the other two subclasses to restore the old behaviour.

ryanheise:
The code is not really copying a file, it is reading an asset via `AssetBundle.load` which doesn't return a file (even if the asset may internally be stored in a file, that is not exposed by the public APIs). Extracting it to a file makes it easy to pass it to the platform APIs since they don't know how to read from Flutter assets directly.

But it's a good question, since it's not perfectly ideal to have to do this. Another way I could implement it in the future is to make assets available to the platform side via the proxy server which directly reads the bytes out of the asset and serves them over HTTP.

For the PR, I'd be happy to merge but can you apply the changes suggested above first?

ryanheise:
Hi @ceppelli , I am not sure if you are available to make the requested changes, so I will begin doing them now on my local repo. A new release should be out soon after.

ryanheise:
The aforementioned changes are now committed and released 0.5.1. Thanks for bringing the error to my attention!

PR #209, opened 2020-10-13
Add web dependency

creativecreatorormaybenot:
If you look at federated plugins, e.g. https://github.com/creativecreatorormaybenot/wakelock or `video_player`, you will see that they add their `web` dependency to the main plugin in their own Pubspec.

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.4+1
+
+* Add web dependency.
+
 ## 0.5.4
 
 * Use audio_session 0.0.9.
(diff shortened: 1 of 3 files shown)


ryanheise:
Thanks, yes I had been meaning to do this soon after bootstrapping the change to the federated plugin model but forgot about it.

LGTM and merged :+1:

ryanheise:
The web dependency should also be removed from the example's pubspec, though (will test later).

creativecreatorormaybenot:
@ryanheise Thanks a lot for the quick response!

So far, I think that your audio plugins are outstanding and I will probably be using them extensively in the near future. If I find something to contribute, I will make sure to do that 👌

ryanheise:
Thanks for the compliment :-) I've also tested the endorsement of `just_audio_web` and it is working well.

Issue #208, opened 2020-10-13
Spaces in local file paths don't work on iOS


gaseline:
**Which API doesn't behave as documented, and how does it misbehave?**
Calling `setFilePath`  with a path that contains spaces does not work, specified file cannot be found. Even though `File(path).exists()` on the same path returns true.

**Minimal reproduction project**
The example

**To Reproduce (i.e. user steps, not code)**
1. Run the app
2. Load a file path the contains spaces, such as absolute path to "Application Support" folder
3. Observe the error

**Error messages**

```
AVPlayerItemStatusFailed
PlatformException -- The requested URL was not found on this server
```

**Expected behavior**
Any allowed character in directory or file names should correctly transfer to the audio player.

**Screenshots**
Not applicable.

**Desktop (please complete the following information):**
 - OS: MacOS 10.15.3
 - Browser: Not applicable

**Smartphone (please complete the following information):**
 - Device: iPhone 7 & iPad Air 1
 - OS: iOS 14 & iOS 12

**Flutter SDK version**
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, v1.17.5, on Mac OS X 10.15.3 19D76, locale en-US)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 12.0.1)
[✓] Android Studio (version 4.0)
[✓] IntelliJ IDEA Community Edition (version 2018.3.6)
[✓] VS Code (version 1.46.0)
[!] Connected device
    ! No devices available

! Doctor found issues in 1 category.
```

**Additional context**
The path seems to be encoded twice. Logging the uri from `UriAudioSource.m` > `initWithId` > under `file://` shows a path that has its spaces transformed into "%2520". It looks like the "%" from "%20" is encoded again. I've double checked the function input for `setFilePath` and this doesn't contain any encoding, it happens in the plugin.

Replacing this line seems to solve the issue
```
_playerItem = [[IndexedPlayerItem alloc] initWithURL:[NSURL fileURLWithPath:[_uri substringFromIndex:7]]];
to
_playerItem = [[IndexedPlayerItem alloc] initWithURL:[NSURL fileURLWithPath:[[_uri stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding] substringFromIndex:7]]];
```

ryanheise:
[...]

PR #214, opened 2020-10-19
Spaces in local file paths don't work on iOS

hsangtini:
https://github.com/ryanheise/just_audio/issues/208


ryanheise:
Awesome, thank you for your contribution!

PR #216, opened 2020-10-19
Remove debug print statements

creativecreatorormaybenot:
## Description

It looks like you forgot to remove some debug print statements from the plugin code.

---

This PR removes these debug print statements.  

### `.gitignore`

Additionally, I added the proper `.gitignore` file to the platform interface directory (copied from the main plugin directory). This is because Git added the IntelliJ files.

## Problem

These `print` statements are somewhat unacceptable to me because they litter the logs with out of context print statements, especially the web print statements are spammy.

I think the best approach is removing them completely.

## Considerations

I am not the first person to think of removing print statements from package production code.  
You can use the [`avoid_print` linter rule](https://dart-lang.github.io/linter/lints/avoid_print.html) to have the analyzer take care of warning you 👍


ryanheise:
That is perfectly understandable, especially for the web. The logging of method calls is certainly something I forgot to remove. The print statement for the stacktrace was intentional since prior to the move to the federated plugin model, the stack trace used to get lost when forwarding the error through the platform channel. However, this is no longer the case, and so I agree with all of your changes.

The only thing is that your diff contains a number of other changes (e.g. to .gitignore) that are not related to debug print statements, and it would be best if you can remove those changes in this PR before I merge it.

creativecreatorormaybenot:
@ryanheise Thanks a lot for taking a look!

The `.gitignore` is actually the only other change. I added it in order to not have other change (`.idea` files for example). What do you think?

ryanheise:
My intention is to eventually define a set of common gitignore patterns at the root level, and will delete .gitignore files in subdirectories. You can choose which files to include/exclude when staging a commit, although in this case you can undo those changes and I'll do a squash merge.

creativecreatorormaybenot:
@ryanheise Done 👍

ryanheise:
Thanks! Merged :-)

PR #234, opened 2020-11-02
setSkipSilence and volumeBoost on android.

subhash279:
I implemented the above for Android. It seems more involved iOS side. So someone with more ObjectiveC experience can probably take it up in future.
Please let me know if anything else is needed.
Please close the older PR I submitted which is not in sync with latest changes.

### DIFF
--- a/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -83,10 +83,10 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Aud
     private MediaSource mediaSource;
     private Integer currentIndex;
     private float _speed = 1;
-	private boolean _skipSilence = false;
-	private boolean _volumeBoostEnabled = false;
-	private int _volumeBoostGainMB = 0;
-	private LoudnessEnhancer loudness;
+    private boolean _skipSilence = false;
+    private boolean _volumeBoostEnabled = false;
+    private int _volumeBoostGainMB = 0;
+    private LoudnessEnhancer loudness;
     private Map<LoopingMediaSource, MediaSource> loopingChildren = new HashMap<>();
     private Map<LoopingMediaSource, Integer> loopingCounts = new HashMap<>();
     private final Handler handler = new Handler();
@@ -148,9 +148,9 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Aud
         } else {
             this.audioSessionId = audioSessionId;
         }
-        if(_volumeBoostEnabled) {
-			setVolumeBoost(true, _volumeBoostGainMB);
-		}
+        if (_volumeBoostEnabled) {
+						setVolumeBoost(true, _volumeBoostGainMB);
+				}
     }
 
     @Override
@@ -324,13 +324,13 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Aud
                 result.success(new HashMap<String, Object>());
                 break;
             case "setSkipSilence":
-				setSkipSilence((Boolean) request.get("enabled"));
-				result.success(null);
-				break;
-			case "setVolumeBoost":
-				setVolumeBoost((Boolean) request.get("enabled"), (Integer) request.get("gainmB"));
-				result.success(null);
-				break;
+				        setSkipSilence((Boolean) request.get("enabled"));
+				        result.success(null);
+				        break;
+						case "setVolumeBoost":
+							setVolumeBoost((Boolean) request.get("enabled"), (Integer) request.get("gainmB"));
+							result.success(null);
+							break;
             case "setLoopMode":
                 setLoopMode((Integer) request.get("loopMode"));
                 result.success(new HashMap<String, Object>());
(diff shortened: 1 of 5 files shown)


ryanheise:
Hi @subhash279 thanks for that. Before I merge, would you mind also using consistent formatting with the existing code?

Note that all files now use spaces instead of tabs (since cc2b10834bebc94b1757576d654da11ea4f49774), and also the use of spaces after "if", again to be consistent with the surrounding code. Though, beware of using auto-format, too, since that does not preserve the existing formatting on lines unrelated to this PR.

Regarding naming conventions, I would like to follow [Effective Dart](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words) on the capitalisation of acronyms, so a name like `gainmB` would need some adjustment, though I see how its difficult to comply. Their assumption was that we could easily understand what the acronym meant if it were written with case like this: `gainMb`. But of course MB stands for different things depending on the casing so mB is a case-sensitive acronym. Perhaps we need to spell out the full word? e.g. `gainMillibels`?

Let me know if you have a different preference or suggestion.

subhash279:
@ryanheise thanks for the info. I made the formatting changes and gainMillibels name change as needed.

ryanheise:
Thanks, @subhash279 , looks good. I will integrate this after first fixing a separate critical bug so that I can get that out first.

ryanheise:
The other bug fix has now landed, so I will look at this soon. Fortunately it didn't force any merge conflict.

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -80,6 +82,11 @@
     private Integer audioSessionId;
     private MediaSource mediaSource;
     private Integer currentIndex;
+    private float _speed = 1;
+    private boolean _skipSilence = false;
+    private boolean _volumeBoostEnabled = false;
+    private int _volumeBoostGainMB = 0;
```
ryanheise, line 88: To follow the convention set by the rest of the file, the leading underscores should be removed.

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -313,6 +323,14 @@ public void onMethodCall(final MethodCall call, final Result result) {
                 setSpeed((float) ((double) ((Double) request.get("speed"))));
                 result.success(new HashMap<String, Object>());
                 break;
+            case "setSkipSilence":
+                setSkipSilence((Boolean) request.get("enabled"));
+                result.success(null);
```
ryanheise, line 328: Following the style of the rest of the file, these should return a JSON-encoded instance of the appropriate return type, in this case `SetSkipSilenceResponse`. The same applies to `setVolumeBoost` below.

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -660,10 +678,31 @@ public void setVolume(final float volume) {
     }
 
     public void setSpeed(final float speed) {
-        player.setPlaybackParameters(new PlaybackParameters(speed));
+        _speed = speed;
+        if (_skipSilence) {
+            player.setPlaybackParameters(new PlaybackParameters(_speed, 1, _skipSilence));
```
ryanheise, line 683: Is there an issue with using this line unconditionally? (i.e. it seems to universally handle both the if/else cases)

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -660,10 +678,31 @@ public void setVolume(final float volume) {
...
         broadcastPlaybackEvent();
     }
 
+    public void setSkipSilence(final boolean skipSilence) {
+        _skipSilence = skipSilence;
+        player.setPlaybackParameters(new PlaybackParameters(_speed, 1, skipSilence));
+    }
+
+    public void setVolumeBoost(final boolean enabled, final int gainmB) {
+        if (android.os.Build.VERSION.SDK_INT >= 19 && audioSessionId != null) {
```
ryanheise, line 696: You imported `android.os.Build` so this doesn't need to be fully qualified.

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -660,10 +678,31 @@ public void setVolume(final float volume) {
...
+            _volumeBoostEnabled = enabled;
+            _volumeBoostGainMB = gainmB;
+            Log.e(TAG, "setVolumeBoost in android 2 : " + enabled);
+            loudness = new LoudnessEnhancer(audioSessionId);
+            loudness.setEnabled(enabled);
+            loudness.setTargetGain(gainmB);
```
ryanheise, line 702: Question: how many AudioEffects do we want to support? It may be worth designing the API considering this bigger picture, and also iOS.

just_audio/darwin/Classes/AudioPlayer.m
```diff
@@ -97,6 +97,9 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
         } else if ([@"setVolume" isEqualToString:call.method]) {
             [self setVolume:(float)[request[@"volume"] doubleValue]];
             result(@{});
+        } else if ([@"setSkipSilence" isEqualToString:call.method]) {
+            /// TODO on iOS side; Seems more involved, so someone with ObjectiveC experience might look at it.
+            result(nil);
```
ryanheise, line 102: Same as for the Android implementation, this should return an encoding of the response object. It seems you are missing the corresponding method call handler for `setVolumeBoost` (although it could well be better to have that as a separate PR once we figure out how to handle audio effects in general).

just_audio/lib/just_audio.dart
```diff
@@ -584,6 +584,23 @@ class AudioPlayer {
...
+  }
+
+  /// Sets volume boost option on exoPlayer to true/false. Works only on Android for now.
+  /// gainMillibels is the value in millibels you want to amplify the volume by.
+  Future<void> setVolumeBoost(bool enabled, {int gainMillibels = 2000}) async {
+    if (_disposed) return;
+    enabled ??= false;
+    if (Platform.isAndroid) {
+      await (await _platform).setVolumeBoost(SetVolumeBoostRequest(enabled: enabled, gainMillibels: gainMillibels));
+    }
```
ryanheise, line 601: I would rather unconditionally pass this request through to the platform and let it decide what to do with it.

ryanheise:
HI @subhash279 , I've left some comments on your code. I think the main one is that the volume boost API design probably needs more thought before we commit to anything, so it would probably be best to separate your volume boost code out into a separate branch. The skip silence code is relatively uncontroversial and that would allow me to merge it sooner.

subhash279:
@ryanheise thanks for the changes. i will go through them and tick them off by this weekend. Regarding VolumeBoost, can I just pass it unconditionally to iOS as well as you said in the last comment. If I have to split it into a different branch, it will be a lot more work and I may not have time soon for that. Once I pass it unconditionally we can come back to that later to see if anyone wants to contribute on iOS side and then decide on the changes.

ryanheise:
My overall thought is that the VolumeBoost part will require careful consideration if we look at what other audio effects we want to support. If we're supporting all of them, then I wouldn't want to have a separate method for each one, I might instead want to have a separate class with an inheritance hierarchy, for example. But this is not something I can decide right away in this pull request, so that's why I suggested removing the VolumeBoost code from this PR and just including the setSkipSilence code.

The skipSilence code is relatively uncontroversial so I can approve that more quickly as is, but as long as the VolumeBoost code is in there, I won't be able to merge it until we've taken the time to fully think through the API design which may take a lot longer. If you'd rather keep both of these features as part of the same pull request, it will just take longer to get merged.

subhash279:
thanks for the details. But do you think because VolumeBoost is such a common feature for podcast apps, it should still be exposed as a separate method for devs to use. So in future even if more complex audio effects are added, the internals can be changed but the usage by devs wouldn't change as it is still a very useful method and it might be better to keep it separate from other audio effects(in terms of method exposure to devs). The internal methods can be figured out later if I just leave them empty method now on iOS side. On Android side, exo player considers volumeBoost as a feature on its own and separate from other audio effects as well.

ryanheise:
> On Android side, exo player considers volumeBoost as a feature on its own and separate from other audio effects as well.

Which API specifically are you referring to?

subhash279:
LoudnessEnhancer was used to achieve this effect on Android. 
https://developer.android.com/reference/android/media/audiofx/LoudnessEnhancer

ryanheise:
The LoudnessEnhancer is not "a feature on its own and separate from the other audio effects", though, it's a subclass of AudioEffect. Is there an example in the Android APIs where volume boost is a separate method rather than a subclass of generalised audio effects?

ryanheise:
I have just merged your implementation of `setSkipSilence` into the `dev` branch with the proposed changes mentioned in my code review above, and updated it to null safety, added a state stream, and integrated it into just_audio's platform lifecycle.

I have kept the volume boost code in the Java file while not exposing the API to Dart yet until I settle on a generic API for audio effects. This will happen after the iOS implementation is converted to an AVAudioSession-based one which should give me some idea of an appropriate way to design this generic API.

Thanks for contributing, and hopefully I will get around to the other half of this (volume boost) before too long.

PR #297, opened 2021-02-03
Fix SequenceState's currentSource & currentIndex when length is zero

smkhalsa:
When a `ConcatenatingAudioSource` has a length of 0 (such as after calling `clear()`), accessing the `SequenceState. currentSource` getter throws an index out of range exception.

Furthermore, `SequenceState`'s docs state that `If [sequence.length] is 0, then [currentIndex] is also 0.`. However, `currentIndex` is currently getting set to `-1` when `sequence.length` is 0.

This PR fixes both issues.

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -129,7 +129,7 @@ class AudioPlayer {
       (sequence, shuffleIndices, currentIndex, shuffleModeEnabled, loopMode) {
         if (sequence == null) return null;
         if (currentIndex == null) currentIndex = 0;
-        currentIndex = min(sequence.length - 1, max(0, currentIndex));
+        currentIndex = max(min(sequence.length - 1, max(0, currentIndex)), 0);
         return SequenceState(
           sequence,
           currentIndex,
@@ -1310,7 +1310,8 @@ class SequenceState {
       this.shuffleModeEnabled, this.loopMode);
 
   /// The current source in the sequence.
-  IndexedAudioSource get currentSource => sequence[currentIndex];
+  IndexedAudioSource get currentSource =>
+      sequence.isEmpty ? null : sequence[currentIndex];
 
   /// The effective sequence. This is equivalent to [sequence]. If
   /// [shuffleModeEnabled] is true, this is modulated by [shuffleIndices].


ryanheise:
Thanks! I'll need to add a unit test for this.

Part of me thinks that it may actually make more sense for `currentIndex` to be `-1` in this case, or maybe `null`, but yes according to the current documentation it is a bug and your solution looks reasonable.

Note that I have already migrated the package to null safety (see the `nnbd` branch), and I haven't yet decided on whether I want to now maintain two separate versions of the plugin or just make all new developments on the `nnbd` branch. Let me think a bit on it, but I'll probably merge your PR on `master` and then try to merge that into `nnbd`.

smkhalsa:
> Part of me thinks that it may actually make more sense for currentIndex to be -1 in this case

I agree, `-1` is clearer as it disambiguates what a `currentIndex` of `0` means. Since it's specifically mentioned in the docs, I wasn't sure if it might break things elsewhere, but if not, I'd advocate for leaving it as `-1`.

> Note that I have already migrated the package to null safety (see the nnbd branch), and I haven't yet decided on whether I want to now maintain two separate versions of the plugin or just make all new developments on the nnbd branch. Let me think a bit on it, but I'll probably merge your PR on master and then try to merge that into nnbd.

I actually tried to use the nnbd branch, but there were too many incompatible dependency issues that weren't trivial to resolve, so I switched back to master. So selfishly, I'd prefer maintaining both for a little while longer as the ecosystem catches up with the migration.

ryanheise:
I've just merged your PR into master. One more thing I noticed is that `get currentSource` may also throw an exception if `sequence` is `null`, which it may be if no audio source has been set yet. I'll fix that in a followup commit, and then merge this over to `nnbd`.

Thanks again for your contribution!

ryanheise:
Regarding the null `sequence`, looking in context, the the code you fixed was inside `SequenceState`, where `sequence` can't be `null`, so there's no problem after all.

smkhalsa:
> Regarding the null `sequence`, looking in context, the the code you fixed was inside `SequenceState`, where `sequence` can't be `null`, so there's no problem after all.

Hmm. It seems like the issue is when sequence is not `null` but empty (such as after calling `ConcatenatingAudioSource.clear()`.

ryanheise:
Understood. The null sequence thing was not actually a problem, just something I mistakenly thought would be a problem, but only because I had misunderstood which part of the code I was looking at in the context diff.

In any case, this has been released in 0.6.10. The nnbd release will have to wait until tomorrow.

Out of curiosity, which of just_audio's nnbd dependencies caused an issue for you? I noticed today an issue with the crypto dependency, so I'll try to fix that tomorrow by using a range constraint.

smkhalsa:
> Out of curiosity, which of just_audio's nnbd dependencies caused an issue for you? I noticed today an issue with the crypto dependency, so I'll try to fix that tomorrow by using a range constraint.

Yes, `crypto` and `uuid` both caused problems, but I was able to get them working with dependency overrides. However, the main issue is with `rxdart`. Their nnbd release changes the API of `ValueStream` which breaks code from another dependency.

ryanheise:
I see... This is surely going to be a long and bumpy road!

Issue #309, opened 2021-02-15
Fatal Exception: com.google.android.exoplayer2.IllegalSeekPositionException


DenisShakinov:
**Which API doesn't behave as documented, and how does it misbehave?**
We've noticed a few random crashes in prod version of our app (Android). Here is stacktrace from firebase console:
```
Fatal Exception: com.google.android.exoplayer2.IllegalSeekPositionException
       at com.google.android.exoplayer2.ExoPlayerImpl.seekTo(ExoPlayerImpl.java:596)
       at com.google.android.exoplayer2.SimpleExoPlayer.seekTo(SimpleExoPlayer.java:1661)
       at com.ryanheise.just_audio.AudioPlayer.onPlayerError(AudioPlayer.java:271)
```

**Minimal reproduction project**
We are unable to reproduce such an issue on our available test devices so steps to reproduce are unclear

**To Reproduce (i.e. user steps, not code)**
(n/a)

**Error messages**
(n/a)

**Expected behavior**
App doesn't crash on any player error.

**Screenshots**
(n/a)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser (n/a)

**Smartphone (please complete the following information):**
 - Device: OPPO CPH1827
 - OS: Android 7.1.1

**Flutter SDK version**
```
[✓] Flutter (Channel stable, 1.22.6, on macOS 11.0.1 20B50 darwin-x64, locale
    en-BY)
 
[✓] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[✓] Xcode - develop for iOS and macOS (Xcode 12.4)
[!] Android Studio (version 4.1)
    ✗ Flutter plugin not installed; this adds Flutter specific functionality.
    ✗ Dart plugin not installed; this adds Dart specific functionality.
[✓] VS Code (version 1.53.2)
[✓] Connected device (2 available)
```

**Additional context**
I quess it may be some android api version or vendor specific issues (Operating systems: 96% Android 7, Devices: 69% OPPO, 25% LENOVO).
We don't do anything special with the library, just createPlayer -> setAsset -> play use case.

AliHoseinpoor:
I have the same problem.
In crashes section in google play console i have this error
![Uploading IMG_20210312_113400.jpg…]()

ryanheise:
Hi @AliHoseinpoor Can you confirm #310 fixes the issue?

[...]

PR #310, opened 2021-02-15
Fix IllegalSeekPositionException crash on Android

DenisShakinov:
Please see #309

### DIFF
--- a/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -265,10 +265,13 @@ public class AudioPlayer implements MethodCallHandler, Player.EventListener, Aud
         errorCount++;
         if (player.hasNext() && currentIndex != null && errorCount <= 5) {
             int nextIndex = currentIndex + 1;
-            // TODO: pass in initial position here.
-            player.setMediaSource(mediaSource);
-            player.prepare();
-            player.seekTo(nextIndex, 0);
+            Timeline timeline = player.getCurrentTimeline();
+            if (nextIndex >= 0 && (timeline.isEmpty() || nextIndex < timeline.getWindowCount())) {
+                // TODO: pass in initial position here.
+                player.setMediaSource(mediaSource);
+                player.prepare();
+                player.seekTo(nextIndex, 0);
+            }
         }
     }
 


just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -265,10 +265,13 @@ public void onPlayerError(ExoPlaybackException error) {
         errorCount++;
         if (player.hasNext() && currentIndex != null && errorCount <= 5) {
             int nextIndex = currentIndex + 1;
-            // TODO: pass in initial position here.
-            player.setMediaSource(mediaSource);
-            player.prepare();
-            player.seekTo(nextIndex, 0);
+            Timeline timeline = player.getCurrentTimeline();
+            if (nextIndex >= 0 && (timeline.isEmpty() || nextIndex < timeline.getWindowCount())) {
```
ryanheise, line 269: Some of these conditions look a bit surprising:

* How can `nextIndex >= 0` be violated? For `nextIndex` to be -1, `currentIndex` would have to be -2 which would indicate a problem elsewhere.
* What is the scenario where `player.hasNext()` AND `timeline.isEmpty()`? If `timeline.isEmpty()`, what does `player.seekTo(nextIndex, 0)` do?

If all is reasonable, it would be good to add a comment above this condition to explain these surprises.

DenisShakinov: yes it does look quit surprising, but this is `onPlayerError` method triggered so anything can happen here imho. Please take a look at this code from exoplayer:
https://github.com/google/ExoPlayer/blob/4b1e0fa9fc643f96ab5bbf4cec9aa7a79e4bc789/library/core/src/main/java/com/google/android/exoplayer2/ExoPlayerImpl.java#L610
I don't think that problem is in `nextIndex < 0` or `timeline.isEmpty()`. I guess the problem is that `nextIndex >= timeline.getWindowCount()` somehow.
Btw could you please provide a purpose of seeking player to beginning on every player error?

ryanheise: I guess that is why the condition looks puzzling. In context, I don't think all of the terms of that condition are appropriate, because we have just computed the next index which should never be less than zero. It seems at least that condition could be removed, but maybe you can confirm it through debugging to check `nextIndex` is never less than zero. What do you think?

I guess the real mystery is why `hasNext` returns true and yet `nextIndex` then seems to be going out of range.

> Btw could you please provide a purpose of seeking player to beginning on every player error?

I believe this was to emulate the iOS behaviour where if you have a playlist and one item fails, it automatically advances to the next item.

DenisShakinov: Although i cannot debug it (issue is not reproducible in normal circumstances) `nextIndex` seems fine for me. I guess the problem in `timeline`, maybe it has been reset to default values somehow right after player error.

ryanheise: That's also my take. If we look at the ExoPlayer code you referenced, it looks like that `windowIndex` condition was only there to guard against an out of bound parameter value passed in by the client, but in our case `nextIndex` is not a parameter and we can guarantee it is within bounds. Can we remove the `nextIndex` condition?

ryanheise: I'm also not sure about the use of the `timeline.isEmpty()` condition. If the timeline is empty, it means there is no "next" to play, and I think we should just do nothing. So just:

```dart
if (nextIndex < timeline.getWindowCount()) {
```

ryanheise:
I've merged your fix with my own proposed changes added. Fingers crossed!

I've also published a new release 0.7.1 with the bug fix.

PR #314, opened 2021-02-19
changes for update to 2.13.1

MichealReed:
Feel free to send this PR to a different branch if there's somewhere else you'd prefer. I have a branch in my fork to explore the new Dash low latency features, will submit new PRs for anything else I uncover.


just_audio/example/pubspec.lock
```diff
@@ -571,5 +557,5 @@ packages:
     source: hosted
     version: "2.2.1"
 sdks:
-  dart: ">=2.10.0-110 <2.11.0"
+  dart: ">=2.12.0-0.0 <=2.12.0-179.0.dev"
```
ryanheise, line 560: I have a separate `nnbd` branch which opts into null safety, so would rather keep `master` on the lower min SDK for now.

just_audio/example/pubspec.lock
```diff
@@ -35,7 +35,7 @@ packages:
       name: async
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.5.0-nullsafety.1"
+    version: "2.5.0-nullsafety.3"
```
ryanheise, line 38: Are these version bumps coming from flutter stable?

MichealReed:
No idea why, but something auto-updated the pubspec lock and made its way into the commit. Since it's not directly the pubspec file we should be fine, but I went ahead and made a commit that brings it to parity with master.

ryanheise:
That would be good, but it still doesn't seem to be in parity with master.

MichealReed:
https://stackoverflow.com/questions/16136739/in-dart-and-pub-should-i-add-pubspec-lock-to-my-gitignore#16136740

Per this, the lock should probably added to the .gitignore. The latest commit was a copy/paste from master so flutter is auto changing something here.

ryanheise:
Nice, I had always wanted to add the lock files to `.gitignore`. From memory, it was originally there in the plugin template, and then I seem to remember a long time ago after a discussion on flutter github issues that they [decided to remove it](https://github.com/flutter/flutter/pull/24209#issuecomment-438873467) from `.gitignore` for the plugin template.

They still haven't added it back into the template, but I did notice Google's own plugins add it to their `.gitignore` so I'd be happy to do the same.

ryanheise:
I've just made the change to `.gitignore` and removed the lock files - you may need to also remove your copy of the lock files for any you changed.

MichealReed:
aaaand it's gone?

ryanheise:
That's weird, when I view the summary of all commits combined on GitHub, it shows up as adding `just_audio/just_audio/example/pubspec.lock` as a new file. But when I look at your repo, that file doesn't exist...

MichealReed:
Yeah I had to pull the deletion of the file over from here before we attempted it merge or it would've cause basing issues.

just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
```diff
@@ -134,7 +134,6 @@ private void startWatchingBuffer() {
         handler.post(bufferWatcher);
     }
 
-    @Override
     public void onAudioSessionId(int audioSessionId) {
```
ryanheise, line 137: The removal of `@Override` looked suspicious as that is there to alert you if the method you're overriding is not in the superclass/interface (in this case `AudioListener`. It looks like it was renamed - can you rename it to `onAudioSessionIdChanged` and add back `@Override`?

MichealReed:
Should be all set now 👍

ryanheise:
Thanks! Looks good apart from the example `pubspec.lock` file still being there. I'm still not quite sure what the situation was with that, but I've merged anyway, and then I added one more commit on my end to remove the file again.

PR #317, opened 2021-02-23
Add Error state to ProcessingState

MichealReed:
I had a dire need for an error state in my application so I thought I would submit a PR here as well. This is verified working on Android, although feel free to comment if there are any more places I should put it to catch all. I did my best to implement iOS and Darwin but I have not been able to test with an Apple device. There may be a few redundant or missed spots on the Apple side as well, so feel free to call those out.


ryanheise:
I am not sure if this is the right approach. You can currently listen to error events on the streams like this:

```dart
playbackEventStream.listen(listener, onError: ...);
```

I think in the case of an unrecoverable error, it makes sense for the "processing" state to enter `idle`, and we don't need to combine other purposes, while for recoverable errors, some other non-error transition may make more sense. The fact that the error occurred could still be detected by listening to the stream.

MichealReed:
No worries here, thanks for the error catching example. I was facing an issue where it would get suck under a loading processing state and never shift to idle so I thought I would solve with an error state.

ryanheise:
Just looking at my last comment it looks like my last sentence was incomplete - not sure what it was originally going to say :-)

But as to your scenario, I'm not sure if the state transitions currently work in the intended way as described, particularly on iOS, so I'll need to do some more work there.

PR #318, opened 2021-02-23
supports cronet

MichealReed:
We should be using cronet for streaming use cases, it will fail and respond much faster than OkHttp. I am not sure what we would have to do to support Apple with this, but I know it is possible. Exoplayer does make this easier on android though from what I can tell. This will also give us support for HTTP3/QUIC if the server supports, but Cronet falls back to http 1.1 fine in my tests.

### DIFF
--- a/just_audio/android/build.gradle
+++ b/just_audio/android/build.gradle
@@ -44,4 +44,6 @@ dependencies {
     implementation 'com.google.android.exoplayer:exoplayer-dash:2.13.1'
     implementation 'com.google.android.exoplayer:exoplayer-hls:2.13.1'
     implementation 'com.google.android.exoplayer:exoplayer-smoothstreaming:2.13.1'
+    implementation 'com.google.android.exoplayer:extension-cronet:2.13.1'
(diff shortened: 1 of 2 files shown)


MichealReed:
https://medium.com/the-react-native-log/using-cronet-in-your-mobile-app-7dda3a89c132

May be able to use the approach here, not familiar enough with Apple native development to make a call.

ryanheise:
This looks interesting. Would it break HTTP 0.9 streams? Would it best be implemented as an option?

MichealReed:
It should still support 0.9, 1.0, 1.1, 2.0, and 3. QUIC should be revealed as an option, but I feel cronet is safer and better to use than okhttp out of the box.

tiloc:
FWIW: Dart 2.18 is now supporting 2 platform specific implementations for http, one optimised for iOS, one optimised for Android (based on cronet). Both still considered a preview.

https://medium.com/dartlang/dart-2-18-f4b3101f146c

ryanheise:
Interesting, so we could implement that for the proxy code (which is written in dart) while also using the ExoPlayer code for the native side.

PR #354, opened 2021-03-24
Destroying all previous players during hot restart

### LINKED ISSUE #134: Dispose native audio player on development hot restart
I noticed that if you're playing a song and you do a ~cold~ hot restart (not hot reload) during development playbackState is lost. Would be nice if it worked like it does with hot reload.

ericmartineau:
Sorry, this PR has some noise - I replaced instances of `new HashMap<String, Object>()` with `emptyMap()`.

This fixes https://github.com/ryanheise/just_audio/issues/134 - at least it's one approach.

When a hot restart happens, the entire flutter runtime is swapped out with a fresh one.  The main isolate is killed, and everything gets wiped clean.  There is no eventing system that allows us to intercept the unloading process.  Any resource cleanup has to happen within the plugin.  However, all the plugins stay registered within the native OS, and they are connected back to the new flutter runtime - they don't even call the plugin registration methods inside native code.

Another complicating factor is that it's possible to have multiple simultaneous AudioPlayer instances.
Finally, I preferred for this to happen automatically, as opposed to adding another method call that developers have to know about.

I added a method to the plugin contract named `disposeAllPlayers`.  It is equivalent to `disposePlayer` except that it has no arguments.  If the native plugin has any registered players, they will each be disposed, and the plugin will reply with the `id` of each disposed player.

I opted to use a global variable on the dart side, because a hot restart will cause the variable to "lose" it's value, which seemed like a pretty reliable way to detect when a fresh initialization was occurring.  I added a call to `disposeAllPlayers` before the first player was registered, and forced any other pending `AudioPlayer` registrations to wait until `disposeAllPlayers` was complete.


ryanheise:
Thanks for the PR! Although just_audio can run multiple player instances in different isolates, so this global variable approach will not work. Maybe you could extend that logic to check all running isolates on the Dart side. You could also theoretically implement all of this on the Dart side by maintaining a Map of instances and sharing that between isolates, then you wouldn't need to change the platform interface or implement the backend several times for each platform.

Also, there have been new FlutterEngine lifecycle methods introduced and I am hoping that actually this can now be solved on the Android side natively (not sure about iOS).

 There are some new lifecycle events in the engine that might be worth investigating,

ericmartineau:
@ryanheise Ugh, you're right.  The first audio player in the second isolate will kill all the players from the first.

I'll look into the lifecycle events you mentioned, but in the meantime, what do you think about adding a global static method, something like `AudioPlayer.destroyAllPlayers()` that could be invoked by the developer explicitly during application initialization?  I hate to change the API since this is really only for development.

ericmartineau:
Two other possible solutions:

Keep the PR as is, but only run the `killAllPlayers` from the main program isolate.  This would not work if the `AudioPlayer` instances are exclusively initialized from child isolates.

or

Implement something similar to what you proposed above, but instead of maintaining a map of player instances, we maintain a global (across all isolates) initialization flag.  The key is to run the `destroyAllPlayers` one time for each hot restart of the flutter application, regardless of how many isolates are spawned.

ddfreiling:
This would be nice to land. Can be quite difficult to work iteratively with `just_audio` right now.

ryanheise:
I will close this as I have now implemented it via FlutterEngine callbacks on the `dev` branch. Unfortunately the API I needed was deprecated but hopefully before it gets deleted there will be an alternative API I can switch to.

Issue #356, opened 2021-03-27
Pending timers during flutter tests


minhqdao:
**Which API doesn't behave as documented, and how does it misbehave?**
When an instance of `AudioPlayer()` is created the usual way, a `!timersPending` error is obtained during `flutter test` although `_audioPlayer`is properly disposed.

**Minimal reproduction project**
The example

**To Reproduce**
1. Remove the `expect` method from `widget_test.dart`
2. Run `flutter test`

**Error messages**

```
══╡ EXCEPTION CAUGHT BY FLUTTER TEST FRAMEWORK ╞════════════════════════════════════════════════════
The following assertion was thrown running a test:
A Timer is still pending even after the widget tree was disposed.
'package:flutter_test/src/binding.dart':
Failed assertion: line 1245 pos 12: '!timersPending'

When the exception was thrown, this was the stack:
#2      AutomatedTestWidgetsFlutterBinding._verifyInvariants (package:flutter_test/src/binding.dart:1245:12)
#3      TestWidgetsFlutterBinding._runTestBody (package:flutter_test/src/binding.dart:806:7)
<asynchronous suspension>
(elided 2 frames from class _AssertionError)
...

The test description was:
  Verify Platform version
```

**Expected behavior**
There should be no pending timers after the audio player was disposed.

**Desktop:**
 - Run `flutter test` in your terminal

**Flutter SDK version**
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel dev, 2.1.0-12.1.pre, on macOS 11.2.3 20D91 darwin-arm,
    locale de-DE)
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
[✓] Xcode - develop for iOS and macOS
[✓] Chrome - develop for the web
[✓] Android Studio (version 4.1)
[✓] IntelliJ IDEA Community Edition (version 2020.3.3)
[✓] VS Code (version 1.54.3)
[✓] Connected device (2 available)

• No issues found!
```

**Additional context**
Alternative way to reproduce: Create an instance of `AudioPlayer()` in any project, call `_audioPlayer.dispose()` in the state's dispose method and run `flutter test`.

ryanheise:
Thanks, @minhqdao . I use a timer for `positionStream` so maybe that's it. Perhaps we just need to add a brief delay before the test finishes so that the `positionStream` has enough time to release all of its momentum.

ryanheise:
[...]

PR #362, opened 2021-04-02
Fix timers

minhqdao:
There were two `!timersPending` issues found during `flutter test`. The first one could be resolved by removing one of the `Timer` instantiations, which was deemed redundant. In this course it became necessary to declare `currentTimer` nullable. To do justice to the doc description, the stream now only emits while the player is playing.

Refactoring a function within `_setPlatformActive` then settled the second error.

The test was also adjusted to find a reasonable widget.

Attempted fix to https://github.com/ryanheise/just_audio/issues/356

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -494,7 +494,7 @@ class AudioPlayer {
       return s;
     }
 
-    Timer currentTimer;
+    Timer? currentTimer;
     StreamSubscription? durationSubscription;
     StreamSubscription? playbackEventSubscription;
     void yieldPosition(Timer timer) {
@@ -515,9 +515,8 @@ class AudioPlayer {
       controller.add(position);
     }
 
-    currentTimer = Timer.periodic(step(), yieldPosition);
     durationSubscription = durationStream.listen((duration) {
-      currentTimer.cancel();
+      currentTimer?.cancel();
       currentTimer = Timer.periodic(step(), yieldPosition);
     }, onError: (Object e, StackTrace stackTrace) {});
     playbackEventSubscription = playbackEventStream.listen((event) {
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -1003,7 +1003,8 @@ class AudioPlayer {
     final currentIndex = this.currentIndex;
     final audioSource = _audioSource;
     final durationCompleter = Completer<Duration?>();
-    _platform = Future<AudioPlayerPlatform>(() async {
+
+    Future<AudioPlayerPlatform> setPlatform() async {
       _playbackEventSubscription?.cancel();
       if (!force) {
         final oldPlatform = await oldPlatformFuture!;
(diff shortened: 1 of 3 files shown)


just_audio/lib/just_audio.dart
```diff
@@ -512,12 +512,13 @@ class AudioPlayer {
         controller.close();
         return;
       }
-      controller.add(position);
+      if (playing) {
+        controller.add(position);
+      }
```
ryanheise, line 517: I understand the idea of this, but what if the user seeks while not playing? Currently the stream over-eagerly emits position events, and then `.distinct()` is used to cancel them out. So that guarantees the correct position updates, but is not the most efficient. So you're onto something here in terms of efficiency, but we will need some test cases for paused position changes based on seeking within a track and skipping between tracks.

just_audio/lib/just_audio.dart
```diff
@@ -1002,7 +1003,8 @@ class AudioPlayer {
     final currentIndex = this.currentIndex;
     final audioSource = _audioSource;
     final durationCompleter = Completer<Duration?>();
-    _platform = Future<AudioPlayerPlatform>(() async {
+
```
ryanheise, line 1006: Can you explain what this change does?

minhqdao: I figured out that this function was the cause for one of the timer errors. The problem was gone after refactoring it. I wish I could give you a better "explanation".

ryanheise: On the surface both ways of writing it seem equivalent, but I am guessing that your version is guaranteed to execute the synchronous portion of the block in the same event cycle.

just_audio/lib/just_audio.dart
minhqdao: Yeah, I checked that and saw that user handling is still taken into account by the `playbackEventStream`. However, I didn't exactly know that `.distinct()` filters out redundant events. I'll add some tests then.

minhqdao:
Will add one more test for switching between audio sources.

ryanheise:
So it turned out that seeking while paused works?

minhqdao:
Yes, it works. I added some more expects to make sure, but you can test it the example app, too. There is no behavioral change because taps in the paused state are handled by the `playbackEventStream`.

ryanheise:
Although play/pause will directly add to `playbackEventStream`, the question is really whether `positionStream` will pick those changes up. It doesn't seem like that would happen only from reading the code, though, so I'm still a bit puzzled how this could work.

ryanheise:
Ah, OK. This bit is what handles it:

```dart
    playbackEventSubscription = playbackEventStream.listen((event) {
      controller.add(position);
    }, onError: (Object e, StackTrace stackTrace) {});
```

ryanheise:
I've merged this into the `minhqdao-fix_timers` branch while I prepare some further changes for the release, but thank you so much for your contribution!

minhqdao:
Glad I could help!

Issue #47, opened 2020-02-27
Playing a remote file and cache the file at the same time


ReaganRealones:
Any suggestion on how I could play a remote files and store the files in cache for faster playing next time? One option is to use [Flutter cache manager](https://pub.dev/packages/flutter_cache_manager) to download it too  when I play it but this would mean downloading twice i.e one from the remote url (for the audioplayer) and one to store in cache for next time. 

How possible is it to play the remote file while caching it for faster playing the next time?

ryanheise:
I'm not sure of any easy way to do this. Some podcast players make it a binary choice: You either stream it, or you download it in advance.

BruceZhang1993:
This feature seems great. The flutter_cache_manager plugin's getFile method will return a Stream of [FileInfo](https://pub.dev/documentation/flutter_cache_manager/latest/flutter_cache_manager/FileInfo-class.html). I wonder whether it's possible to achieve this feature.

lkho:
> The flutter_cache_manager plugin's getFile method will return a Stream of `FileInfo`

The `FileInfo` stream only provides the progress of the download (while the file is still downloading), but not a data stream. You need to wait for the whole file to be downloaded to get a real file path.

BTW, if you really want to play a URL while it is still being downloaded, there is a [`CacheDataSourceFactory`](https://exoplayer.dev/downloading-media.html) in ExoPlayer, which might serve this purpose. However, this is platform side implementation and not controlled by Flutter. If it was to be implemented in this library, it would be a huge feature and requires a lot of platform communications.

ryanheise:
I would eventually like to replace the current `setUrl`/`setXYZ` API by a `setDataSource` API which would make it a lot easier to integrate with custom and other ExoPlayer data sources. There are various other motivations for wanting to do this including gapless playback. So I would say this idea is certainly on the table, but not in the short term.

ryanheise:
I'll leave this link here for later as a guide for how this could be done on the iOS side:

https://vombat.tumblr.com/post/86294492874/caching-audio-streamed-using-avplayer

sachaarbonel:
I was reading the rxdart doc when I came across this [feature](https://pub.dev/documentation/rxdart/latest/rx/SwitchIfEmptyStreamTransformer-class.html) that could become handy for this feature request

ryanheise:
I am not sure since the downloading actually happens on the platform side of the plugin.

[...]

PR #401, opened 2021-05-12
Add caching error handling

kmod-midori:
As discussed in #47, return a 500 in case of errors. Based on my own private implementation of `LockCachingAudioSource`.

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -2268,6 +2268,8 @@ class LockCachingAudioSource extends StreamAudioSource {
             contentType: mimeType,
             stream: response,
           ));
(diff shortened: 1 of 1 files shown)


ryanheise:
Nice! At a first glance, it all looks good. Probably the documentation of `StreamAudioSource.request` should mention that implementations can throw an exception to feed an error to the consumer of that stream, i.e. the audio decoding pipeline.

chengyuhui:
The documentation has been adjusted.

ryanheise:
I have (finally) merged this onto the `dev` branch - thanks for contributing! I made some minor changes including clearing `_requests` on an error, and adding a download progress stream. I've also added another example, `lib/example_caching.dart` to demonstrate this.

PR #463, opened 2021-07-29
Fix iOS12.4 `LockCachingAudioSource` fetch data exception

akindone:


just_audio/lib/just_audio.dart
```diff
@@ -2707,13 +2707,15 @@ class LockCachingAudioSource extends StreamAudioSource {
       for (var cacheResponse in inProgressResponses) {
         if (_progress >= cacheResponse.end) {
           // We've received enough data to fulfill the byte range request.
-          cacheResponse.controller.add(
-              data.sublist(0, data.length - (_progress - cacheResponse.end)));
+          var subEnd = (data.length - (_progress - cacheResponse.end))
+              .clamp(0, data.length);
+          cacheResponse.controller.add(data.sublist(0, subEnd));
           cacheResponse.controller.close();
```
ryanheise, line 2713: Looking at the documentation for [clamp](https://api.dart.dev/stable/2.13.4/dart-core/num/clamp.html), it gives no guarantee that the result will be an `int`, because it doesn't have generic types, just `num`. Maybe it would be safer to use `min` and `max` instead?

akindone:
> Can we leave this change out of the pull request?

removed

ryanheise:
Thanks for doing that, I will try to take a look later today and do some testing.

ryanheise:
Apologies for the delay on releasing it, but it is now merged and released as 0.9.6. Thank you for the contribution!

PR #480, opened 2021-08-09
Fix equality

cedvdb:
Similar to this PR on flutter fire https://github.com/FirebaseExtended/flutterfire/pull/5443/files

The problem with using dynamics here is that it makes it hard to mock AudioPlayer


ryanheise:
Thanks for the contribution! I have pulled your branch into my own so that I can make some further related changes (checking `runtimeType` and also using `hashValues`). I'll merge after checking up on some other features I may want to include in the next release.

ryanheise:
This is now published as 0.9.7.

PR #519, opened 2021-09-19
Windows Support

bdlukaa:
#103

All the essential features are supported. Here's the Feature list compared to the other platforms:

| Feature                        | Android   | iOS     | macOS   | Web     | Windows |
| -------                        | :-------: | :-----: | :-----: | :-----: | :-----: |
| read from URL                  | ✅        | ✅      | ✅      | ✅      | ✅    |
| read from file                 | ✅        | ✅      | ✅      | ✅      | ✅    |
| read from asset                | ✅        | ✅      | ✅      | ✅      | ✅    |
| read from byte stream          | ✅        | ✅      | ✅      | ✅      |          |
| request headers                | ✅        | ✅      | ✅      |         |        |
| DASH                           | ✅        |          |         |         |  ✅       |
| HLS                            | ✅        | ✅      | ✅      |         |    ✅    |
| ICY metadata                   | ✅        | ✅      | ✅      |         |        |
| buffer status/position         | ✅        | ✅      | ✅      | ✅      |      |
| play/pause/seek                | ✅        | ✅      | ✅      | ✅      | ✅    |
| set volume/speed               | ✅        | ✅      | ✅      | ✅      | ✅    |
| clip audio                     | ✅        | ✅      | ✅      | ✅      |        |
| playlists                      | ✅        | ✅      | ✅      | ✅      | ✅    |
| looping/shuffling              | ✅        | ✅      | ✅      | ✅      |   ✅   |
| compose audio                  | ✅        | ✅      | ✅      | ✅      |       |
| gapless playback               | ✅        | ✅      | ✅      |         | ✅    |
| report player errors           | ✅        | ✅      | ✅      | ✅      |       |
| handle phonecall interruptions | ✅        | ✅      |         |         |        |
| buffering/loading options      | ✅        | ✅      | ✅      |         |        |
| set pitch                      | ✅        |         |         |         |         |
| skip silence                   | ✅        |         |         |         |         |
| equalizer                      | ✅        |         |         |         |         |
| volume boost                   | ✅        |         |         |         |         |

This week I will add the features that were not implemented yet, and also partial Linux support.


just_audio/example/lib/example_playlist.dart
```diff
@@ -23,18 +23,18 @@ class MyApp extends StatefulWidget {
...
+    //   end: Duration(seconds: 90),
+    //   child: AudioSource.uri(Uri.parse(
+    //       "https://s3.amazonaws.com/scifri-episodes/scifri20181123-episode.mp3")),
+    //   tag: AudioMetadata(
+    //     album: "Science Friday",
+    //     title: "A Salute To Head-Scratching Science (30 seconds)",
+    //     artwork:
+    //         "https://media.wnyc.org/i/1400/1400/l/80/1/ScienceFriday_WNYCStudios_1400.jpg",
+    //   ),
+    // ),
```
bdlukaa, line 37: I commented this out because this audio source is not supported yet

just_audio/example/lib/example_playlist.dart
```diff
@@ -328,26 +328,27 @@ class ControlButtons extends StatelessWidget {
...
                 onPressed: player.pause,
               );
+            } else if (processingState == ProcessingState.loading ||
+                processingState == ProcessingState.buffering) {
+              return Container(
+                margin: EdgeInsets.all(8.0),
+                width: 64.0,
+                height: 64.0,
+                child: CircularProgressIndicator(),
+              );
```
bdlukaa, line 351: I moved this down here because the indicator was spinning forever.

just_audio/lib/just_audio.dart
```diff
@@ -1254,6 +1255,11 @@ class AudioPlayer {
         }
       }
       if (_disposed) return _platform;
+
+      if (_isLibWinMedia()) {
+        LibWinMediaJustAudioPlugin.registerWith();
+      }
```
bdlukaa, line 1261: Override the current platform interface with `LibWinMediaJustAudioPlugin` whenever we are running on Windows or Linux!

just_audio/example/lib/example_playlist.dart
ryanheise: I would suggest leaving this example unchanged as a sort of litmus test for the correct behaviour. But perhaps you could include an `example` subdirectory within the just_audio_libwinmedia repo. Regarding the bug itself, this would indicate that `processingState` never reaches `ready`.

just_audio/lib/just_audio.dart
ryanheise: For now let's do something similar to `just_audio_background` where the dependency is not automatically endorsed but rather the app developer adds the dependency themselves and calls the `registerWith` method themselves (the latter step hopefully being automated in a future version of Flutter, because it seems they did fix this in a recent PR).

ryanheise:
Hi @bdlukaa could you also include a link to `just_audio_libwinmedia` in the README.md along with the feature check boxes?

just_audio/example/lib/example_playlist.dart
bdlukaa: Indeed I never called the ready enum!

just_audio/lib/just_audio.dart
bdlukaa: But we don't use platform code here, only FFI. There's no way for flutter to register Windows or Linux automatically.

We can take the official path_provider plugin as example:

https://github.com/flutter/plugins/blob/master/packages/path_provider/path_provider/lib/path_provider.dart#L44-L61

Windows and Linux requires manual registration.

ryanheise: Yes but as I mentioned above there was recent work on flutter to rectify this. See https://github.com/flutter/flutter/issues/52267

just_audio/example/analysis_options.yaml
```diff
@@ -0,0 +1,29 @@
+# This file configures the analyzer, which statically analyzes Dart code to
```
ryanheise, line 1: This doesn't need to be part of this PR

just_audio/lib/just_audio.dart
```diff
@@ -9,6 +9,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:just_audio_platform_interface/just_audio_platform_interface.dart';
+import 'package:just_audio_libwinmedia/just_audio_libwinmedia.dart';
```
ryanheise, line 12: There should not be any dependencies from the plugin frontend plugin to the platform implementations.

just_audio/example/analysis_options.yaml
bdlukaa: Sorry! It was automatically generated by flutter

just_audio/lib/just_audio.dart
ryanheise: Actually, looking at the diff for this file, none of the changes are needed. Maybe the only necessary change is the change to the `README.md` file with the supported features and with instructions linking to the just_audio_libwinmedia page for windows support, and the example can be included within the just_audio_libwinmedia repository in the `example` subdirectory (see just_audio_background for an example).

ryanheise:
Hi @bdlukaa I'm going to sleep for the next 8 hours FYI so I won't be able to reply until tomorrow. But remember that I'd like to follow the model of just_audio_background where it is not automatically enabled, but an app developer can add it by adding a dependency alongside just_audio. The just_audio_background example is also defined within the just_audio_background repository. This has advantages in that it is faster to make changes to just_audio_background independently including to its example, and its dependencies (e.g. if they may conflict with just_audio's own example's dependencies.) You should find the same advantages with just_audio_libwinmedia because you'll be able to fix bugs in the plugin and update the example in the same repo (and potentially the same commit).

bdlukaa:
> I'd like to follow the model of just_audio_background where it is not automatically enabled

Then there's no point on this PR. This PR adds support for Windows the same way all the other four platforms are supported, with no additional setup for the dev.

ryanheise:
Apologies if I caused earlier confusion, but yes the plan is to only update the README in this PR and yes to also require additional setup for the developer to use a Windows platform implementation (which is honestly not a big obstacle for the developer since I already do the same for just_audio_background). There are several reasons:

1. I mentioned earlier a Flutter PR that fixes one of the problems with the federated plugin model. Part of what they are working on fixing is to allow having a default implementation for a platform and then being able to override it. For now, that is not possible, and so we can't actually have a default (since that would make it impossible for anyone to override the default and create their own platform implementation, either open source or proprietary -- and some companies need to do this because they have their own DRM solution that they want to use instead of the public implementation). I may be aware of this limitation now, but the reason why the Android, iOS, macOS and web implementations are all endorsed as default implementations is partly because when I originally wrote all of those implementations and then split them up via the federated plugin model, I didn't know that the model had this bug that would not allow overriding the default implementation. This is bad because I actually want to create an alternative implementation of both the iOS and Android implementations and those efforts are currently blocked because of this Flutter limitation.
2. When it comes to endorsing a platform implementation, assuming Flutter can sort out the above bug, I think it would make sense to also not have any endorsed implementation until it stabilises and supports at least the features in `example/lib/example_playlist.dart` as a baseline. It's not something that needs to be done immediately from the very first release.
3. In terms of the example, as mentioned above, I think it will be easier for maintenance to have the windows example in the `just_audio_libwinmedia` repository so that commits can be synchronised across both the library implementation and the example.

bdlukaa:
@ryanheise Hello! I think now this PR is ready for review.

All I did was:

- Added the `windows` folder to the example app
- Added `just_audio_libwinmedia` as the default package to windows and linux (works partially on linux)

just_audio/lib/just_audio.dart
```diff
@@ -1254,6 +1254,7 @@ class AudioPlayer {
         }
       }
       if (_disposed) return _platform;
+
```
ryanheise, line 1257: Better to leave this file unchanged.

just_audio/example/lib/main.dart
```diff
@@ -202,4 +202,4 @@ class ControlButtons extends StatelessWidget {
       ],
     );
   }
-}
+}
```
ryanheise, line 205: This file should be unchanged.

ryanheise:
Hi @bdlukaa LGTM. I'll merge in the morning. I've left a couple of comments about some parts of the diff where the only thing changed in the file was whitespace - those changes are best left out of the diff. I'm happy to also fix this myself after merging, but I'll check back in the morning before merging.

just_audio/example/lib/main.dart
```diff
@@ -202,4 +202,4 @@ class ControlButtons extends StatelessWidget {
...
```
ryanheise, line 205: The difference here is not a trailing space, but rather a UNIX-style text format where every line is terminated by a newline character.

bdlukaa: which means?

ryanheise:
I've merged into the `feature/windows` branch and made one change as described above which is to make the windows/linux implementation something that can be manually added via an additional dependency:

```yaml
dependencies:
  just_audio: any # substitute version number
  just_audio_libwinmedia: any # substitute version number
```

Since I don't have Windows, would you be able to confirm whether this branch works?

I also added a `linux` directory to the example, although I'm currently not able to test that either despite having Linux due to https://github.com/harmonoid/libwinmedia/issues/6

bdlukaa:
@ryanheise It does work!

I'd ask for you to not publish it now because I'm waiting for https://github.com/harmonoid/libwinmedia/pull/7 to land on pub.dev

PR #540, opened 2021-10-07
Update windows and linux

bdlukaa:
Now that Linux is not freezing anymore, we can add it to the README and mark as supported. But some users have reported bug when `preload: true`

Added error reporting for Windows.

| Feature                        | Android | iOS | macOS | Web | Windows | Linux |
| ------------------------------ | :-----: | :-: | :---: | :-: | :-----: | :---: |
| read from URL                  |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| read from file                 |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| read from asset                |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| read from byte stream          |   ✅    | ✅  |  ✅   | ✅  |         |       |
| request headers                |   ✅    | ✅  |  ✅   |     |         |       |
| DASH                           |   ✅    |     |       |     |   ✅    |       |
| HLS                            |   ✅    | ✅  |  ✅   |     |   ✅    |       |
| ICY metadata                   |   ✅    | ✅  |  ✅   |     |         |       |
| buffer status/position         |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| play/pause/seek                |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| set volume/speed               |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| clip audio                     |   ✅    | ✅  |  ✅   | ✅  |         |
| playlists                      |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| looping/shuffling              |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| compose audio                  |   ✅    | ✅  |  ✅   | ✅  |         |
| gapless playback               |   ✅    | ✅  |  ✅   |     |   ✅    |       |
| report player errors           |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
| handle phonecall interruptions |   ✅    | ✅  |       |     |         |       |
| buffering/loading options      |   ✅    | ✅  |  ✅   |     |         |       |
| set pitch                      |   ✅    |     |       |     |         |       |
| skip silence                   |   ✅    |     |       |     |         |       |
| equalizer                      |   ✅    |     |       |     |         |       |
| volume boost                   |   ✅    |     |       |     |         |       |

### DIFF
--- a/just_audio/README.md
+++ b/just_audio/README.md
@@ -34,31 +34,31 @@ This project is supported by the amazing open source community of GitHub contrib
 
 ## Features
 
-| Feature                        | Android   | iOS     | macOS   | Web     | Windows |
-| -------                        | :-------: | :-----: | :-----: | :-----: | :-----: |
-| read from URL                  | ✅        | ✅      | ✅      | ✅      | ✅      |
-| read from file                 | ✅        | ✅      | ✅      | ✅      | ✅      |
-| read from asset                | ✅        | ✅      | ✅      | ✅      | ✅      |
-| read from byte stream          | ✅        | ✅      | ✅      | ✅      |         |
-| request headers                | ✅        | ✅      | ✅      |         |         |
-| DASH                           | ✅        |         |         |         | ✅      |
-| HLS                            | ✅        | ✅      | ✅      |         | ✅      |
-| ICY metadata                   | ✅        | ✅      | ✅      |         |         |
-| buffer status/position         | ✅        | ✅      | ✅      | ✅      | ✅      |
-| play/pause/seek                | ✅        | ✅      | ✅      | ✅      | ✅      |
-| set volume/speed               | ✅        | ✅      | ✅      | ✅      | ✅      |
-| clip audio                     | ✅        | ✅      | ✅      | ✅      |         |
-| playlists                      | ✅        | ✅      | ✅      | ✅      | ✅      |
-| looping/shuffling              | ✅        | ✅      | ✅      | ✅      | ✅      |
-| compose audio                  | ✅        | ✅      | ✅      | ✅      |         |
-| gapless playback               | ✅        | ✅      | ✅      |         | ✅      |
-| report player errors           | ✅        | ✅      | ✅      | ✅      |         |
-| handle phonecall interruptions | ✅        | ✅      |         |         |         |
-| buffering/loading options      | ✅        | ✅      | ✅      |         |         |
-| set pitch                      | ✅        |         |         |         |         |
-| skip silence                   | ✅        |         |         |         |         |
-| equalizer                      | ✅        |         |         |         |         |
-| volume boost                   | ✅        |         |         |         |         |
+| Feature                        | Android | iOS | macOS | Web | Windows | Linux |
+| ------------------------------ | :-----: | :-: | :---: | :-: | :-----: | :---: |
+| read from URL                  |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
+| read from file                 |   ✅    | ✅  |  ✅   | ✅  |   ✅    | ✅    |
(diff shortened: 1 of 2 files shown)


alexmercerind:
@ryanheise I believe these API changes have their made their way into `just_audio_libwinmedia` (along `libwinmedia`)'s version `0.0.4` (you are on `0.0.3`).

ryanheise:
pub versioning: https://dart.dev/tools/pub/versioning

ryanheise:
I would just point out that it is not difficult at all to adhere to the pub.dev semantic versioning convention for breaking changes. When you implement a breaking change, instead of bumping from 0.0.6 to 0.0.7, you bump from 0.0.6 to 0.1.0. It's what we all have to do to be good citizens on pub.dev, and to ensure that we don't break other people's apps in CI or hosted packages. It's not as through we have all adopted our own rules, we follow the same contract so that the pub versioning constraints actually work. Yes, this PR is clearly intended to update and not fix, but since libwinmedia's incorrect semantic version propagated through the pub system, it has caused a breakage in another package which this PR now fixes.

In the case of a flutter favorite, which this is, things are a bit more strict, and one of the rules is that I am not allowed to include any dependencies in just_audio that do not follow best practices (because things like this can happen) or the flutter favorite status can be revoked.

I hope you understand, and I hope that you will consider following Dart's semantic versioning convention that I linked above in future, but if not, I would suggest a safer option that solves the dependency issue is for @bdlukaa to host the Linux/Windows examples in the just_audio_libwinaudio repository so that if something like this happens again, the damage is limited to that repository and doesn't risk the flutter favorite status of this plugin. (This is actually how I set things up for just_audio_background knowing that that separation would make it easier to modify the dependencies of each repository independently, and it is ideally the model for other federated platform implementations too, at least before they become stable.)

ryanheise:
That's true for semantic versioning in general, but Dart has its own conventions (see link above). In particular, for any version pre 1.0.0, we treat the middle version number to signify a breaking change. So in 0.x.y+z, increment z for a change that does not affect the public API, increment y for a change that adds a feature or fixes a bug in a backwards compatible way, and increment x if a breaking change was made. Once a package becomes stable, increment the 0 to 1, and then different versioning rules apply.

ryanheise:
See this section:

> Although semantic versioning doesn’t promise any compatibility between versions prior to 1.0.0, the Dart community convention is to treat those versions semantically as well. The interpretation of each number is just shifted down one slot: going from 0.1.2 to 0.2.0 indicates a breaking change, going to 0.1.3 indicates a new feature, and going to 0.1.2+1 indicates a change that doesn’t affect the public API. For simplicity’s sake, avoid using + after the version reaches 1.0.0.

ryanheise:
I've just merged and published this in just_audio 0.9.13. I will later check what was going on with my own issue running on Linux, although just after quickly checking again now I'm getting a blank window rather than the loading indicator, which I'm guessing might be due to the same freeze issue discovered earlier. But I'll be able to confirm that later with some debugging.

bdlukaa:
@ryanheise try to set `preload: false`

bdlukaa:
@ryanheise It seems like the states are never being triggered, but you still can play the audio after calling load

ryanheise:
The latest commit updates the playlist example to set `preload: false` when on Linux, and with a bit of luck I actually got the example to work one time. I think it's still a bit hit and miss, though. When it worked, it only worked after hitting play (silence), then pause, then play (the second time was audible).

That said, I think it's a good foundation to build on. Hopefully people who need Linux will now start experimenting with it and may be able to help contribute to libwinmedia via pull requests.

bdlukaa:
@ryanheise Hopefully! I'm glad this was merged and able to contribute to this.

PR #546, opened 2021-10-11
Add ios FFT & visualizer widget

Eittipat:
This is "just work" code,  Optimization & Memory leaked checking are needed.

### DIFF
--- a/just_audio/darwin/Classes/AudioPlayer.m
+++ b/just_audio/darwin/Classes/AudioPlayer.m
@@ -10,12 +10,14 @@
 #import <AVFoundation/AVFoundation.h>
 #import <stdlib.h>
 #include <TargetConditionals.h>
+#include "AndroidFFT.h"
 
 typedef struct JATapStorage {
     void *self;
     Float64 samplingRate;
     int captureSize;
     void *waveform;
+    void *fft;
 } JATapStorage;
 
 // TODO: Check for and report invalid state transitions.
@@ -429,6 +431,7 @@ static void initTap(MTAudioProcessingTapRef tap, void *clientInfo, void **tapSto
     storage->self = clientInfo;
     storage->captureSize = self.visualizerCaptureSize;
     storage->waveform = calloc(1, self.visualizerCaptureSize);
+    storage->fft = calloc(1, self.visualizerCaptureSize);
     *tapStorageOut = storage;
 }
 
@@ -451,7 +454,7 @@ static void processTap(MTAudioProcessingTapRef tap, CMItemCount frameCount, MTAu
 
     UInt8 *waveform = (UInt8 *)storage->waveform;
     int captureSize = storage->captureSize;
-    
+
     AudioBuffer *buffer = &bufferListInOut->mBuffers[0];
     int bufferSize = buffer->mDataByteSize;
     // TODO: Handle interleaved channels.
@@ -467,15 +470,21 @@ static void processTap(MTAudioProcessingTapRef tap, CMItemCount frameCount, MTAu
         else if (unsignedSample < 0) unsignedSample = 0;
         waveform[i] = (UInt8)unsignedSample;
     }
-
+    
     // TODO: Take captureRate into account. Maybe let the main thread
     // periodically take samples, and we provide them here in a CMSimpleQueue.
     AudioPlayer *self = (__bridge AudioPlayer *)(storage->self);
     // NOTE: Apple recommends to NOT allocate any memory in the tap process function.
     // TODO: Check impact on performance and memory.
+    
+    // FFT
+    UInt8 *fftMag = (UInt8*)storage->fft;
+    [AndroidFFT doFft:fftMag :waveform :captureSize];
+    
     dispatch_async(dispatch_get_main_queue(), ^{
         NSData *data = [NSData dataWithBytes:(void *)waveform length:captureSize];
-        [self broadcastVisualizerCapture:data samplingRate:(int)(storage->samplingRate)];
+        NSData *data2 = [NSData dataWithBytes:(void *)fftMag length:captureSize];
+        [self broadcastVisualizerCapture:data :data2 samplingRate:(int)(storage->samplingRate)];
     });
 }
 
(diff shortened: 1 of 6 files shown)


just_audio/darwin/Classes/AudioPlayer.m
```diff
@@ -462,16 +470,32 @@ static void processTap(MTAudioProcessingTapRef tap, CMItemCount frameCount, MTAu
...
+    
     // TODO: Take captureRate into account. Maybe let the main thread
     // periodically take samples, and we provide them here in a CMSimpleQueue.
     AudioPlayer *self = (__bridge AudioPlayer *)(storage->self);
     // NOTE: Apple recommends to NOT allocate any memory in the tap process function.
     // TODO: Check impact on performance and memory.
+    
+    // FFT
+    // TODO: Implement FFT using bit operation to speed up
+    fft(captureSize, realData,imageData);
```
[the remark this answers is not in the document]
Eittipat: Sorry for the bad English,

Yes and I've already put it on the bottom of FFTHelper.c, in the comment, just in case if you need it.

There is a scaling problem. Because apple-based FFT is floating-point and we need **int8** for sake of android compatibility.  Thus, the scaling process is needed and I don't know how to do it correctly. So I decided to use integer-based FFT.
However, at the bottom of the line, I have to scale int to int8. Either way, I can't get rid of the scaling problem.

I have to choose between scaling float to int8 or int to int8. I think the floating range is too large (too many possible numbers), so I decided to use integer-based FFT. Another advantage is that usage of integer FFT is more portable. No need to initiate or destroy DSP engine. and If you worry about its performance, the integer-based FFT can be done with + and * which can transform into a bitshift operation which is really fast.

I think there may be a possibility to implement int8 FFT so we don't have to scale the result. But I fade difficulties on overflow problem so I stop making it.

I can switch to floating FFT (apple-based) if you think it provides a better result.

Cheers,

just_audio/ios/Classes/FFTHelper.m
```diff
@@ -0,0 +1,67 @@
+//
+//  FFTHelper.c
+//  just_audio
+//
+//  Created by Eittipat on 10/10/2564 BE.
+//
+//  Original source codes
+//  - https://developer.android.com/reference/android/media/audiofx/Visualizer#getFft(byte[])
+//  - http://www.myuiviews.com/2016/03/04/visualizing-audio-frequency-spectrum-on-ios-via-accelerate-vdsp-fast-fourier-transform.html
```
ryanheise, line 9: I've been stuck on just the the legal side of things with copying this code, and the code you have linked to is not actually licensed (i.e. no permission is granted to copy it.)

I will try to contact the original author and ask him for permission, otherwise it would be necessary to write this clean.

ryanheise: I know how you feel, but copyright is a little tricky. It is better to get permission (which I'm in the process of getting), unless you were to write it clean without looking at the other code. How it gets complicated is that copyright protects not the idea but the expression of the idea. So for example, it could be argued that it would be very unlikely for you to have chosen the variable names that you did were not not for following the original code. And as for whether you could simply rename all the variables, the issue here is that there is now a historical record of how you arrived at the end result, which started with a copy and then making superficial changes (in the court of law they would also look at whether it could be proved that you had access to the original work - although let me just say as a disclaimer that IANAL). So anyway, yes to ensure that everything is OK it is always best to cross the t's and dot the i's in legal matters, and companies who link just_audio into their apps would also want be able to trust that all code is legitimately licensed.

Personally when I read a tutorial, I like to first check if the code has a license, and if not, I will read the tutorial only to learn, but then try to write my own code based only on the understanding I gained from the article rather than by constantly looking to the next line of code in their example before writing my own. In any case, it should be hopefully resolved by contacting the original author and just getting permission.

Eittipat:
Hello,

To get rid of the copyright problem. I switched to android version FFT (took from AOSP). 
This code should give exactly the same result & performance as android.

I also rewrote apple-based FFT from scratch and used different functions to compute FFT.
I hope this can avoid any license problems.

Eittipat:
Hello, @ryanheise. 

Any update on this. Don't hesitate to tell me if you need anything.

ryanheise:
Hi @Eittipat 

While looking at the AOSP code, did you manage to find the answer to what sort of scaling Android actually does?

Regarding the license issue, I've already had to add an Apache license in the past to incorporate the ExoPlayer dependency, but it's important this is also made known to Flutter by adding it to the top level LICENSE file. I've already added the Apache license there but you'd just need to modify the sentence that introduces the license to say that it not only applies to ExoPlayer but this new code you're incorporating.

I'm a bit sick at the moment, but I'll see what I can do to help you get this merged.

Eittipat:
Hello @ryanheise 

Hope you're doing well. I just updated the LICENSE file.

About the scaling algorithm. I just roughly read it.
So  I don't know what it does for now.

ryanheise:
Just wondering, is this intended to be iOS only?

```bash
$ ls -l ios/Classes
-rw-r--r-- 1 ryan ryan 1176 Jan 23 14:53 AndroidFFT.h
-rw-r--r-- 1 ryan ryan 7522 Jan 23 14:53 AndroidFFT.mm
-rw-r--r-- 1 ryan ryan  610 Jan 23 14:53 AudioPlayer.h
lrwxrwxrwx 1 ryan ryan   34 Nov  3 13:14 AudioPlayer.m -> ../../darwin/Classes/AudioPlayer.m
...etc...
```

In all other cases, I have put the `*.m` files in the shared `darwin` directory since the same code typically works on both iOS and macOS, with only the header files needing to be slightly different in each case. If there's no specific reason otherwise, I can make the small change on merge.

Eittipat:
It's should work on any platform. It's pure c implementation. However I only tested on ios.

ryanheise:
Hi @Eittipat I've just merged your PR but am just writing an FFT widget that takes into account the Android-based FFT data layout. Will comment once that's done.

ryanheise:
I've just added a couple of commits to:

1. Symlink your iOS code to support macOS.
2. Implement an FFT visualizer widget.
3. Automatically convert the platform FFT data to signed within the plugin.

Thanks @Eittipat for the awesome work on this PR!

Issue #603, opened 2021-12-07
AirPlay support for locally stored files


samnung:
**Is your feature request related to a problem? Please describe.**
I would like to support playing locally stored MP3 files via AirPlay. In order to enable that I need to set `AVPlayer.allowsExternalPlayback` to `false`. For more info see this SO https://stackoverflow.com/a/69141019/16082219

**Describe the solution you'd like**
I would like to have way to set that property or just_audio setting that property to false for local files.

**Describe alternatives you've considered**

**Additional context**
I've tried setting that property locally and it works. So it is just setting that property and nothing else.

I have no idea why the logic about that property is reversed 🤷 

BTW thank you so much for this AMAZING library. It works really well.

PR #629, opened 2022-01-13
Add setAllowsExternalPlayback

samnung:
Implementation of #603


ryanheise:
Looks like it should work. I'm not set up to be able to test AirPlay so I would appreciate some more testing (although it does seem relatively uncontroversial.)

If anyone out there stumbles across this pull request and can test it, let me know below how it goes.

PR #657, opened 2022-02-08
fix: update loaded audio source position

### LINKED ISSUE #656: Wrong position is reported when load new audio source on iOS
`setAudioSource` reported wrong position if there is already loaded source.
More specifically, last position of former audio source is reported while loading new audio source.

https://github.com/initig/just_audio/blob/fix_ios_wrong_position_report_example/just_audio/example/lib/example_ios_wrong_position.dart

General way
1. load an audio source
2. move position to X
3. load another audio source
4. then position X is reported right after start to load another audio source

Use minimal reproduction project
1. execute example_ios_wrong_position.dart
2. tap audio 1
3. move position to X
4. tap audio 2
5. you can see position moves to X when loading audio 2, after loaded it goes to 0:00

Sample Logs
```
[log] load audio 1 start
[log] load audio 1 done
[log] position: 0:00:00.000000
[log] position: 0:32:15.629000
[log] load audio 2 start
[log] position: 0:00:00.000000
[log] position: 0:32:15.629000 // <- wrong position reported!
[log] load audio 2 done
[log] position: 0:00:00.000000
```

**Expected behavior**
Not report any of former audio source position

 - Device: [Simulator (iPhone13), iPhone SE(2020)]
 - OS: [iOS15(Simulator), iOS15.2(iPhone SE)]

**Flutter SDK version**
```
Flutter (Channel stable, 2.8.1, on macOS 12.1 21C52 darwin-arm, locale
    ja-JP)
```

addie9000:
fixed #656

`updatePosition` in load seems too early. At this moment, it will update former audio source position and report it.
This PR moves `updatePosition` to a later line to report new audio source position.
(If initial position is set, it will be reported right after attached)

### DIFF
--- a/just_audio/darwin/Classes/AudioPlayer.m
+++ b/just_audio/darwin/Classes/AudioPlayer.m
@@ -578,7 +578,6 @@
     }
     _loadResult = result;
     _processingState = loading;
-    [self updatePosition];
     _index = (initialIndex != (id)[NSNull null]) ? [initialIndex intValue] : 0;
     // Remove previous observers
     if (_indexedAudioSources) {
@@ -628,6 +627,7 @@
         [self addItemObservers:source.playerItem];
         source.playerItem.audioSource = source;
     }
+    [self updatePosition];
     [self updateOrder];
     // Set up an empty player
     if (!_player) {


ryanheise:
You need to be careful with this one - please take a look at the previous change that I just made to this file. It's a similar sort of bug but where `_index` hadn't been updated in time before the broadcast, so I moved it up. But your pull request will break my previous fix.

addie9000:
I took a look the previous commit 5bea1121e24b4d80e588c0d9436019a5f6505a3c.
If I understand the problem correctly, index error happens because of using the new _index with old _indexedAudioSources.
But in this PR, new _index is used with new _indexedAudioSources, so the index out of bounds should not happen.

I've tested locally by create 1 audio source and concatenated 2 audio sources with initial index 1.

The way of testing
1. load 1 audio source.
2. load concatenated 2 audio sources.

Results
1. before commit 5bea1121e24b4d80e588c0d9436019a5f6505a3c ->  `Error loading audio source: (9999999) Error in handleMethodCall` happened
2. after commit 5bea1121e24b4d80e588c0d9436019a5f6505a3c  -> no error happened
3. this PR -> no error happened

ryanheise:
Ah, yes I see, I think this looks correct. I will try to test when I get a chance.

ryanheise:
After testing, it looked good to me so it's merged and released as 0.9.19. I did notice a similar issue though when switching to another audio source with an `initialPosition`. Here on iOS it also briefly flips back to position zero before arriving at `initialPosition`. For now I've just left a TODO comment in the code.

PR #696, opened 2022-03-28
Update just_audio_background.dart

flutterbest:


ryanheise:
Thanks, @flutterbest . I'm not sure if this is the right way to do it, though. There are situations where the current index can be undefined (`null`) and that should have a different meaning from zero. The current code is definitely inconsistent, since `nextIndex` and `previndex` are not nullable, instead they return -1 for undefined. I can take a look at that later, but since `nextIndex` and `prevIndex` currently are not nullable, I would say that this `getRelativeIndex` should also return -1 if `index` is undefined.

Thoughts? Do you have a particular reproduction case for context? Normally, a PR would be linked to a bug.

PR #701, opened 2022-04-06
No need to cast [long] parameter to [int] when expected argument is [long]

aam:


ryanheise:
LGTM. I want to combine a few PRs for the next release so probably will tend to this on the weekend.

ryanheise:
This PR is now published.

Issue #572, opened 2021-10-27
Pods issue on macOS Monterey 12.0.1


jonny7737:
**Which API doesn't behave as documented, and how does it misbehave?**
Not an API issue.  build of package example fails.

**Minimal reproduction project**
The example

**To Reproduce (i.e. user steps, not code)**
1. From project directory
2. flutter run
3. build fails

**Error messages**

```
    Resolving dependencies of `Podfile`
      CDN: trunk Relative path: CocoaPods-version.yml exists! Returning local
      because checking is only performed in repo update
    [!] CocoaPods could not find compatible versions for pod "just_audio":
      In Podfile:
        just_audio (from `Flutter/ephemeral/.symlinks/plugins/just_audio/macos`)

    Specs satisfying the `just_audio (from
    `Flutter/ephemeral/.symlinks/plugins/just_audio/macos`)` dependency were
    found, but they required a higher minimum deployment target.
```

**Expected behavior**
Expected example app to run.

**Screenshots**
No screens => no screenshots.

**Desktop (please complete the following information):**
 - OS: macOS 12.0.1
 - None - Desktop application

**Smartphone (please complete the following information):**
 - Device:N/A
 - OS: N/A

**Flutter SDK version**
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 2.5.3, on macOS 12.0.1 21A559 darwin-arm, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0)
[✓] Xcode - develop for iOS and macOS
[✓] Chrome - develop for the web
[✓] Android Studio
[✓] VS Code (version 1.53.0-insider)
[✓] Connected device (2 available)

• No issues found!
```

**Additional context**
I would very much like to try this audio player.

ryanheise:
Hi @jonny7737 any update on this?

ryanheise:
I see, have you tried the official example and does it contain the bug you reported?

jonny7737:
Official example has the same issue:

ld: warning: dylib (/Users/jonny/Downloads/just_audio-master/just_audio/example/build/macos/Build/Products/Debug/just_audio/just_audio.framework/just_audio) was built for newer macOS version (10.12.2) than being linked (10.11)

And the build runs as expected.

FaKenKoala:
Below solution is working for me:

1.  comment the `platform :osx, '10.11'` in macos/Podfile
2. delete Podfile.lock and run pod install:
[...]

PR #735, opened 2022-05-23
Fix for version warning discussed in issue #572

Ruzo:
A tiny change in file /just_audio/macos/just_audio.podspec:

s.platform = :osx, '10.12.2' to s.platform = :osx, '10.11'.


Ruzo:
Fixed that one thing and left all other warnings alone. Let me know if you prefer to find another way to fix this issue instead of bumping down the platform version.

Ruzo (commented):
Fixed that one thing and left all other warnings alone. Let me know if you prefer to find another way to fix this issue instead of bumping down the platform version.

ryanheise:
I assume this one is all sorted now.

PR #797, opened 2022-08-15
Fix User Agent not sent on HTTP redirects

### LINKED ISSUE #796: User Agent is not added to redirected URLs
When configuring the player to use a custom User Agent, the User-Agent header is only sent on the first request. However, if the url is a redirect url (which seems to be quite common on the podcast world), the following requests don't contain the configured user agent.

Use the default project, just change the audio url and configure the user agent (as explained on next step).

1. Configure a custom user agent on the player
2. Play an audio which resolves to a redirect: (https://play.podtrac.com/npr-510282/edge1.pod.npr.org/anon.npr-mp3/npr/pchh/2022/08/20220810_pchh_05aaa4b9-373a-4307-ac0d-5012aad593cb.mp3?awCollectionId=510282&awEpisodeId=1116295815&orgId=1&topicId=1138&aggIds=93568166&d=1544&p=510282&story=1116295815&t=podcast&e=1116295815&size=24717733&ft=pod&f=510282)
3. Check the network requests and check the different redirected requests. Only the first one has the correct user agent.

**Error messages**
N/A

**Expected behavior**
The user agent should also be set on redirected requests.

**Screenshots**
N/a

**Additional context**
Currently the ProxyHttpClient used to set the user agent is doing so by setting the user agent header on the request. However, according to this issue https://github.com/dart-lang/http/issues/268, that header is not added to the redirects. The proper way of doing that is to set the userAgent property on the HttpClient() instance.

I will create a PR to address this issue. Let me know if you need more information.

snipd-mikel:
Fixes #796

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Fix bug in ConcatenatingAudioSource.clear().
 * Fix bug where proxy drains origin faster than it feeds client.
+* Fix bug where User Agent was not set on redirects (@mikel-snipd)
+* Fix bug where StreamAudioSource requests are not closed when proxy is disposed (@mikel-snipd)
 
 ## 0.9.28
 
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3129,13 +3129,15 @@ _ProxyHandler _proxyHandlerForUri(
   Map<String, String>? headers,
   String? userAgent,
 }) {
+  // Keep redirected [Uri] to speed-up requests
+  Uri? redirectedUri;
   Future<void> handler(_ProxyHttpServer server, HttpRequest request) async {
     final client = HttpClient();
 
     if (userAgent != null) {
       client.userAgent = userAgent;
     }
-    final originRequest = await client.getUrl(uri);
+    final originRequest = await client.getUrl(redirectedUri ?? uri);
 
     // Rewrite request headers
     final host = originRequest.headers.value('host');
@@ -3163,6 +3165,9 @@ _ProxyHandler _proxyHandlerForUri(
     // Try to make normal request
     try {
       final originResponse = await originRequest.close();
+      if (originResponse.redirects.isNotEmpty) {
+        redirectedUri = originResponse.redirects.last.location;
+      }
 
       request.response.headers.clear();
       originResponse.headers.forEach((name, value) {
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3103,12 +3103,16 @@ _ProxyHandler _proxyHandlerForSource(StreamAudioSource source) {
     }
 
     final completer = Completer<void>();
-    stream.listen((event) {
+    final subscription = stream.listen((event) {
       request.response.add(event);
     }, onError: (Object e, StackTrace st) {
       source._player?._playbackEventSubject.addError(e, st);
     }, onDone: () {
       completer.complete();
+    });
+
+    request.response.done.then((dynamic value) {
+      subscription.cancel();
     });
 
     await completer.future;
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3137,6 +3137,9 @@ _ProxyHandler _proxyHandlerForUri(
     final host = originRequest.headers.value('host');
     originRequest.headers.clear();
 
+    // Match ExoPlayer's native behavior
+    originRequest.maxRedirects = 20;
+
     // Make sure that we send the userAgent header also on the first request
     if (userAgent != null) {
       originRequest.headers.set(HttpHeaders.userAgentHeader, userAgent);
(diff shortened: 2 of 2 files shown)


jsongenova:
This doesn't seem to work as intended? The initial request now doesn't have the configured user agent but only the redirects.
In `addUriAudioSource` the following lines are deleted but those add the user agent to the initial request:
```
if (source._player?._userAgent != null) {
  headers['user-agent'] = source._player!._userAgent!;
}
```

snipd-mikel:
Hi @jsongenova, thanks for checking. I will take a look. My understanding was that `client.userAgent = userAgent;` will make sure that all the requests are done using the given user agent.

snipd-mikel:
I think this: `originRequest.headers.clear();` might be the issue, which clears all the headers on the first request, including the `user-agent`. Will fix it now

jsongenova:
> I think this: `originRequest.headers.clear();` might be the issue, which clears all the headers on the first request, including the `user-agent`. Will fix it now

Thanks for the fix! I actually stumbled upon something a bit more serious regarding redirects: The default setting for HttpClient's `maxRedirect` property is 5. There are a lot of podcasts that redirect > 5 times, which isn't a problem when not using the proxy but when using the proxy I had to add `originRequest.maxRedirects = 10;` to overcome this problem.

When this happens, eventually `AudioPlayer.setUrl` fails with a Source exception that stems from a socket time out.  It turns out that in the handler supplied by `_proxyHandlerForUri` there's a `catch HttpException` where the max redirect exception gets eaten, so after that nothing happens and after a while it just times out.

The offending URL is: https://arttrk.com/p/KSTA5/chrt.fm/track/B7AB5D/pdst.fm/e/dts.podtrac.com/redirect.mp3/traffic.omny.fm/d/clips/9b7dacdf-a925-4f95-84dc-ac46003451ff/1713c520-edb6-43a3-b1b9-acb8002fdae7/3235ab47-a729-4a73-8158-acb8002fe292/audio.mp3?utm_source=Podcast&in_playlist=58e33a0c-f86b-41c5-a11c-acb8002fdaf5

Which has been taken directly from an RSS feed for 'The Portal' podcast redirects a whopping 7 times. I'm not sure what a sane default value is or what the default is for the ExoPlayer on Android, but it has never failed me so far.

snipd-mikel:
That's a very good point regarding the redirects. Seeing how the podcasts urls are being built I really think 5 is a very low value for `maxRedirects`. I think the limit should probably be increased by default or at least allow to make it configurable. @ryanheise what is your view on this?

ryanheise:
Good point. I suggest we bump `maxRedirects` to 20, as this will match ExoPlayer's native behaviour.

snipd-mikel:
Makes sense. I also pushed that change onto the PR.

ryanheise:
Thanks, @snipd-mikel and @jsongenova . I'm testing it.

snipd-mikel:
@ryanheise  I have made some more tests and found out some worrying behaviour when using the proxy together with `ClippingAudioSource`.
I have a feature where the user can scroll through podcast clips, so I use a lot of `ClippingAudioSources`. When using the proxy, the data consumption quickly grows, **getting to the GB area after a couple of items** (it seems that instead of just getting the data for the clip, the full podcast episode is downloaded). 

However, when not using a proxy (not configuring a user agent or headers), the data doesn't grow as much, as it looks like only the needed data is downloaded.

My hypothesis is that when using the proxy, the `HttpServer` somehow doesn't understand when the client (`ExoPlayer` or `AVPlayer`) wants more data or not, so it keeps loading the podcast episode. I still have to do some more investigation around this, to see how the players request data.

* Have you already encountered something like this?
* And also, what is the reason to use a proxy instead of passing the headers to the native layer and configuring on the respective AudioSources?

ryanheise:
Thanks for reporting. May I ask if you are just using the headers feature or are you also using `LockCachingAudioSource`? The source of the eager buffering could be in a different place depending on which.

Using native headers on ExoPlayer is straightforward. AVPlayer has historically been problematic because the API to do so was unofficial, and that was the reason for originally introducing the proxy in the first place. However, with the recent announcement by Apple of `AVURLAssetHTTPUserAgentKey`, we can now reconsider that, at least for that one specific header. If you're using other headers like authorization headers, then AVPlayer does not offer any "public" API to do it, and you're left with either using an undocumented API (and having your app rejected), or using a workaround.

The proxy approach is a common approach to get around this on iOS, and implementing the proxy also ensures consistent behaviour across platforms. But yes, any eager buffering issue will need to be fixed, definitely. Are you able to create a separate issue for this?

snipd-min:
Hi @ryanheise, thanks for the explanation regarding the proxy. We are not using `LockCachingAudioSource` at the moment. Do you have an idea where the eager buffering could be happening? I will create a separate issue on this when I can make a minimal repo to reproduce it.

snipd-mikel:
@ryanheise I just rebased this.

snipd-mikel:
I have added one more improvement to the proxy: After the first request to a given Uri, it will cache the redirected Uri for that handler. This will speed up subsequent request made to that same Uri.

As mentioned on the comments, a lot of podcast hosting companies heavily use redirects to implement analytics, tracking and dynamic ads, between other things. And just the process of resolving the redirects can take up to 5 seconds. So by caching this we can skip the redirect resolve process on the next range request that the native player performs.

ryanheise:
I like it :+1:

 I'm having trouble merging right now since my main development box is having WIFI issues. I hope to have things back up within a couple of days.

ryanheise:
This has now been merged.

PR #864, opened 2022-11-03
Feature: Add `package` param to `AudioPlayer.setAsset()`

### LINKED ISSUE #863: `AudioPlayer.setAsset` method should have `package` parameter
Flutter reads asset paths differently when the asset is in another package, e.g:
Let we have a `my_flutter_app` and a package named `app_assets` and in there we got a mp3 file `assets/compressed/hello.mp3`.
You can't pass it in `my_flutter_app` as `assets/compressed/hello.mp3` because it is in another package, so the app wouldn't be able to load it. The url / path should be: `packages/app_assets/assets/compressed/hello.mp3`

The method could have a `String? package` parameter, then it would check if it's null or else use the original inputted asset path.

```dart
Future<Duration?> setAsset(
  String assetPath, {
  String? package,
  bool preload = true,
  Duration? initialPosition,
}) {
  final keyName = package == null ? assetPath : 'packages/$package/$assetPath';
  return setAudioSource(
    AudioSource.uri(Uri.parse('asset:///$keyName')),
    initialPosition: initialPosition, preload: preload,
  );
}
```

An alternative is to manually set the `packages/package_name/...`.

**Additional context**
The current `AssetImage` and `IconData` and other packages (e.g, lottie) do a similar approach to assets in packages.

ewertonls:
Add package argument to `setAsset` so assets in packages can be loaded.
closes #863

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.9.31
+
+* Add AudioPlayer.setAsset() a package argument (@ewertonls)
+
 ## 0.9.30
 
 * Upgrade ExoPlayer to 2.18.1.
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -692,15 +692,25 @@ class AudioPlayer {
   /// setAudioSource(AudioSource.uri(Uri.parse('asset:///$assetPath')),
   ///     initialPosition: Duration.zero, preload: true);
   /// ```
+  /// The [package] argument must be non-null if an asset from package is being
+  /// fetched.
   ///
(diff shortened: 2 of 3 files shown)


ryanheise:
I'm reviewing through my phone but have you updated the version in pubspec.yaml?

ewertonls:
> I'm reviewing through my phone but have you updated the version in pubspec.yaml?

I didn't push it before, but I've just updated it

ryanheise:
I've now merged it into `minor` and also added tests. Thanks for the contribution!

I also added an `AudioSource.asset()` convenience method since others had also asked for that in the past.

Let me know if I broke anything on your end ;-)

PR #880, opened 2022-11-21
just_audio_background: small performance fix and lint fixes

Zekfad:
perf: `_PlayerAudioHandler#_broadcastState` make translation map const
chore: use Object.hash for `TrackInfo`
chore: mark `TrackInfo` as immutable
~~chore: apply [`discarded_futures`](https://dart-lang.github.io/linter/lints/discarded_futures.html) lint~~
chore: apply [`unnecessary_await_in_return`](https://dart-lang.github.io/linter/lints/unnecessary_await_in_return.html) lint
chore: apply [`avoid_void_async`](https://dart-lang.github.io/linter/lints/avoid_void_async.html) lint

There's also [`avoid_classes_with_only_static_members`](https://dart-lang.github.io/linter/lints/avoid_classes_with_only_static_members.html) lint from Effective Dart, but that would unnecessary be breaking change.

### DIFF
--- a/just_audio_background/lib/just_audio_background.dart
+++ b/just_audio_background/lib/just_audio_background.dart
@@ -238,7 +238,7 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
       playerDataController.stream;
 
   @override
-  Future<LoadResponse> load(LoadRequest request) async =>
+  Future<LoadResponse> load(LoadRequest request) =>
       _playerAudioHandler.customLoad(request);
 
   @override
@@ -289,7 +289,7 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
 
   @override
   Future<ConcatenatingInsertAllResponse> concatenatingInsertAll(
-          ConcatenatingInsertAllRequest request) async =>
+          ConcatenatingInsertAllRequest request) =>
       _playerAudioHandler.customConcatenatingInsertAll(request);
 
   @override
@@ -299,18 +299,18 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
 
   @override
   Future<ConcatenatingMoveResponse> concatenatingMove(
-          ConcatenatingMoveRequest request) async =>
+          ConcatenatingMoveRequest request) =>
       _playerAudioHandler.customConcatenatingMove(request);
 
   @override
   Future<SetAndroidAudioAttributesResponse> setAndroidAudioAttributes(
-          SetAndroidAudioAttributesRequest request) async =>
+          SetAndroidAudioAttributesRequest request) =>
       _playerAudioHandler.customSetAndroidAudioAttributes(request);
 
   @override
   Future<SetAutomaticallyWaitsToMinimizeStallingResponse>
       setAutomaticallyWaitsToMinimizeStalling(
-              SetAutomaticallyWaitsToMinimizeStallingRequest request) async =>
+              SetAutomaticallyWaitsToMinimizeStallingRequest request) =>
           _playerAudioHandler
               .customSetAutomaticallyWaitsToMinimizeStalling(request);
 }
@@ -351,7 +351,7 @@ class _PlayerAudioHandler extends BaseAudioHandler
   List<MediaItem>? get currentQueue => queue.nvalue;
 
   _PlayerAudioHandler(String playerId) {
-    unawaited(_init(playerId));
+    _init(playerId);
   }
 
   Future<void> _init(String playerId) async {
@@ -492,7 +492,7 @@ class _PlayerAudioHandler extends BaseAudioHandler
               SetAutomaticallyWaitsToMinimizeStallingRequest request) async =>
           (await _player).setAutomaticallyWaitsToMinimizeStalling(request);
 
-  Future<void> _updateQueue() async {
+  void _updateQueue() {
     queue.add(sequence.map((source) => source.tag as MediaItem).toList());
   }
 
@@ -597,10 +597,10 @@ class _PlayerAudioHandler extends BaseAudioHandler
   Future<void> rewind() => _seekRelative(-AudioService.config.rewindInterval);
 
   @override
-  Future<void> seekForward(bool begin) async => _seekContinuously(begin, 1);
+  Future<void> seekForward(bool begin) => _seekContinuously(begin, 1);
 
   @override
-  Future<void> seekBackward(bool begin) async => _seekContinuously(begin, -1);
+  Future<void> seekBackward(bool begin) => _seekContinuously(begin, -1);
 
   @override
   Future<void> setRepeatMode(AudioServiceRepeatMode repeatMode) async {
@@ -677,14 +677,14 @@ class _PlayerAudioHandler extends BaseAudioHandler
     }
   }
 
-  Future<void> _broadcastStateIfActive() async {
+  void _broadcastStateIfActive() {
     if (_justAudioEvent.processingState != ProcessingStateMessage.idle) {
       _broadcastState();
     }
   }
 
   /// Broadcasts the current state to all clients.
-  Future<void> _broadcastState() async {
+  void _broadcastState() {
     final controls = [
       if (hasPrevious) MediaControl.skipToPrevious,
       if (_playing) MediaControl.pause else MediaControl.play,
--- a/just_audio_background/lib/just_audio_background.dart
+++ b/just_audio_background/lib/just_audio_background.dart
@@ -238,9 +238,8 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
       playerDataController.stream;
 
   @override
-  Future<LoadResponse> load(LoadRequest request) async {
-    return _playerAudioHandler.customLoad(request);
-  }
+  Future<LoadResponse> load(LoadRequest request) async =>
+      _playerAudioHandler.customLoad(request);
 
   @override
   Future<PlayResponse> play(PlayRequest request) async {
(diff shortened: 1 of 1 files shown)


just_audio_background/lib/just_audio_background.dart
```diff
@@ -358,17 +351,17 @@ class _PlayerAudioHandler extends BaseAudioHandler
...
   Future<void> _init(String playerId) async {
     final player = await _platform.init(InitRequest(id: playerId));
     _playerCompleter.complete(player);
-    final playbackEventMessageStream = player.playbackEventMessageStream;
-    playbackEventMessageStream.listen((event) {
-      _justAudioEvent = event;
-      _broadcastState();
-    });
+    final playbackEventMessageStream = player.playbackEventMessageStream
+      ..listen((event) {
```
ryanheise: Is this for a lint, or is it your personal preference?

just_audio_background/lib/just_audio_background.dart
```diff
@@ -358,17 +351,17 @@ class _PlayerAudioHandler extends BaseAudioHandler
   List<MediaItem>? get currentQueue => queue.nvalue;
 
   _PlayerAudioHandler(String playerId) {
-    _init(playerId);
+    unawaited(_init(playerId));
```
ryanheise: Not a fan of `unawaited_futures`.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -285,41 +284,35 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
...
+      _playerAudioHandler.customSetAndroidAudioAttributes(request);
 
   @override
   Future<SetAutomaticallyWaitsToMinimizeStallingResponse>
       setAutomaticallyWaitsToMinimizeStalling(
-          SetAutomaticallyWaitsToMinimizeStallingRequest request) async {
-    return await _playerAudioHandler
-        .customSetAutomaticallyWaitsToMinimizeStalling(request);
-  }
+              SetAutomaticallyWaitsToMinimizeStallingRequest request) async =>
```
ryanheise: Here and elsewhere, you should remove the `async` keyword as well.

just_audio_background/lib/just_audio_background.dart
Zekfad: This is actually `discarded_futures`, so I can add `.ignore()` instead if you wish, but that would discard possible unhandled future rejection.

just_audio_background/lib/just_audio_background.dart
Zekfad: This is `cascade_invocations` lint.

just_audio_background/lib/just_audio_background.dart
Zekfad: Done

just_audio_background/lib/just_audio_background.dart
Zekfad: I've removed `unawaited` though.

just_audio_background/lib/just_audio_background.dart
ryanheise: If that lint is turned on (which I didn't realise it was), then let's turn it off. Blindly using cascading invocations does not add to clarity.

just_audio_background/lib/just_audio_background.dart
ryanheise: This specific pattern was already part of the existing `unawaited_futures` lint. `discarded_futures` goes beyond that to prevent other patterns which in my opinion is even less desirable. Note that if your original motivation was unhandled exceptions, `unawaited` doesn't address that anyway, it is merely equivalent to the code I already had but just decorates the code to make it clear that what I did was intentional.

On the other hand, if you call `ignore`, that is not the same meaning as what was originally written as you are throwing away exceptions that could previously be caught by `runZonedGuarded`.

I don's see any clear benefits of either of those approaches. Is there an actual problem you are trying to solve? If so, let's discuss that first before the lints.

Zekfad:
@ryanheise I've made all requested changes, could you review please?

just_audio_background/lib/just_audio_background.dart
```diff
@@ -238,9 +238,8 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
...
 
   @override
-  Future<LoadResponse> load(LoadRequest request) async {
-    return _playerAudioHandler.customLoad(request);
-  }
+  Future<LoadResponse> load(LoadRequest request) =>
+      _playerAudioHandler.customLoad(request);
 
   @override
   Future<PlayResponse> play(PlayRequest request) async {
```
ryanheise, line 245: This async can be deleted.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -285,41 +284,35 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
       _playerAudioHandler.customSetShuffleOrder(request);
```
ryanheise, line 284: And this future should be returned.

ryanheise: Actually my local copy already does that. Maybe I'm viewing these diffs on GitHub wrong? Not sure.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -443,20 +436,18 @@ class _PlayerAudioHandler extends BaseAudioHandler
     return LoadResponse(duration: response.duration);
   }
 
-  Future<SetVolumeResponse> customSetVolume(SetVolumeRequest request) async {
-    return await (await _player).setVolume(request);
-  }
+  Future<SetVolumeResponse> customSetVolume(SetVolumeRequest request) async =>
```
ryanheise, line 439: This async can also be deleted.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -443,20 +436,18 @@ class _PlayerAudioHandler extends BaseAudioHandler
...
+      (await _player).setVolume(request);
 
-  Future<SeekResponse> customPlayerSeek(SeekRequest request) async {
-    return await (await _player).seek(request);
-  }
+  Future<SeekResponse> customPlayerSeek(SeekRequest request) async =>
+      (await _player).seek(request);
```
ryanheise, line 443: Same here.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -443,20 +436,18 @@ class _PlayerAudioHandler extends BaseAudioHandler
...
 
   Future<SetShuffleOrderResponse> customSetShuffleOrder(
       SetShuffleOrderRequest request) async {
     _source = request.audioSourceMessage;
     _updateShuffleIndices();
     _broadcastStateIfActive();
-    return await (await _player).setShuffleOrder(SetShuffleOrderRequest(
+    return (await _player).setShuffleOrder(SetShuffleOrderRequest(
```
ryanheise: Does the lint require this change? When a method is declared as `async`, I prefer to await all futures in the body.

just_audio_background/lib/just_audio_background.dart
Zekfad: You may be viewing old files try to show changes from all commits:
image

just_audio_background/lib/just_audio_background.dart
Zekfad: It's not, because of `await player`?

just_audio_background/lib/just_audio_background.dart
Zekfad: Not sure if it's lint, but there's no need to wait for the value to return if its Future, as it makes no difference (I'm not sure though if it's has performance cost, I mean you have to wait and then return, but you can return and wait for the target result instead of middleware function result)

just_audio_background/lib/just_audio_background.dart
Zekfad: Or you want to use `_player.then((p) => p.setVolume())`, that would be a sync call that returns Future?

just_audio_background/lib/just_audio_background.dart
Zekfad: There's `await _audioHandler.play();` next line, so It's not really possible to remove the `async` except for using `.then`?

just_audio_background/lib/just_audio_background.dart
ryanheise: Sorry I missed that one. We need the await and therefore the async.

Zekfad: Got it, then I resolve these 2 comments.

just_audio_background/lib/just_audio_background.dart
ryanheise: Sorry, I was again mixed up with the GitHub viewing experience. This is correct as is.

just_audio_background/lib/just_audio_background.dart
Zekfad: It's actually returned as you can see here:

https://github.com/ryanheise/just_audio/pull/880/files/1368d8b67104865cfe13e5bc31c43b6f8dbba600#diff-c2bbf6ff0f0e12197c04baff3b8347fc40a1293d426d9f46016ce848a621b15fL282-L285

just_audio_background/lib/just_audio_background.dart
ryanheise: Again, my mistake. The inner await requires the async.

just_audio_background/lib/just_audio_background.dart
ryanheise: Almost certainly these would compile the same way. But for consistency, I prefer to always use await inside of an `async` method.

Zekfad: Oh, its actually due to the lint - adding `await` back raises [unnecessary_await_in_return](https://dart-lang.github.io/linter/lints/unnecessary_await_in_return.html) should I ignore and add await back or it's ok as it is now?

ryanheise: It sounds like unnecessary_await_in_return is something I wouldn't want to adopt. Treating return types as lax (might be a future, might not be a future) doesn't sit well with me.

Zekfad: Ok then I'll return `await`.

Zekfad:
@ryanheise changes made, take a look, please.

ryanheise:
I have just pulled in your changes into the `Zekfad-minor` branch and made some further changes. Mainly, if a method is declared as async, it will return the nested value and let the compiler wrap that in a Future. If you don't have any strong objections, I'd be happy to merge those combined changes.

Zekfad:
Looks good to me, thank you for your work!

ryanheise:
And thank you @Zekfad for the contribution. It has now been merged.

PR #884, opened 2022-11-27
Fix bug 871, 'IcMetadata.Info can be stale'

mrelac:
Android platform only: When a new track is selected that has a Metadata header with a metadataInterval of -1 (i.e. C.LENGTH_UNSET), onMetadata() is never called; thus, the stale metadata info from the last track is not cleared. This fix clears the metadata info under this condition.


fedelubax:
@ryanheise any plans to merge this?

ryanheise:
I had plans to change the API altogether to be less tied to ICY metadata, however I think it's reasonable to probably merge this now. But have you tested it yourself and found it to work well?

PR #914, opened 2023-01-05
feat: notification config builder

### LINKED ISSUE #912: just_audio_background: notification builder
It would be very helpful to have customizable notification behavior for `just_audio_background`.

Expose builder for modifying notification config.

Its's possible to add config? But I think it wound add unnecessary complexity.

**Additional context**
Currently notification properties is controlled via this chunk of code:

https://github.com/ryanheise/just_audio/blob/0098b9b38891cc06ed33ff4a7996d65606e3d193/just_audio_background/lib/just_audio_background.dart#L679-L694

So we can make a builder that will receive micro state snapshot (has previous, has next, etc), and use it's result in this method.

I think its overkill to use `audio_service` for this single change, because you would need to reinvent (or copy-paste) 99% of already existing code from `just_audio_background`.

Zekfad:
resolves #912 

adds `NotificationConfig` that contains config
adds `NotificationConfigBuilder` that is used to accure config
adds `PlayerStateSnapshot` that provides access to state to builder

### DIFF
--- a/just_audio_background/lib/just_audio_background.dart
+++ b/just_audio_background/lib/just_audio_background.dart
@@ -11,6 +11,7 @@ export 'package:audio_service/audio_service.dart' show MediaItem;
 
 late SwitchAudioHandler _audioHandler;
 late JustAudioPlatform _platform;
+late NotificationConfigBuilder _notificationConfigBuilder;
 
 /// Provides the [init] method to initialise just_audio for background playback.
 class JustAudioBackground {
@@ -48,6 +49,7 @@ class JustAudioBackground {
     Duration rewindInterval = const Duration(seconds: 10),
     bool preloadArtwork = false,
     Map<String, dynamic>? androidBrowsableRootExtras,
+    NotificationConfigBuilder? notificationConfigBuilder,
   }) async {
     WidgetsFlutterBinding.ensureInitialized();
     await _JustAudioBackgroundPlugin.setup(
@@ -69,6 +71,7 @@ class JustAudioBackground {
       rewindInterval: rewindInterval,
       preloadArtwork: preloadArtwork,
       androidBrowsableRootExtras: androidBrowsableRootExtras,
+      notificationConfigBuilder: notificationConfigBuilder,
     );
   }
 }
@@ -91,7 +94,10 @@ class _JustAudioBackgroundPlugin extends JustAudioPlatform {
     Duration rewindInterval = const Duration(seconds: 10),
     bool preloadArtwork = false,
     Map<String, dynamic>? androidBrowsableRootExtras,
+    NotificationConfigBuilder? notificationConfigBuilder,
   }) async {
+    _notificationConfigBuilder = notificationConfigBuilder
+      ?? _PlayerAudioHandler._defaultNotificationConfigBuilder;
     _platform = JustAudioPlatform.instance;
     JustAudioPlatform.instance = _JustAudioBackgroundPlugin();
     _audioHandler = await AudioService.init(
@@ -165,7 +171,7 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
   late final _PlayerAudioHandler _playerAudioHandler;
 
   _JustAudioPlayer({required String id}) : super(id) {
-    _playerAudioHandler = _PlayerAudioHandler(id);
+    _playerAudioHandler = _PlayerAudioHandler._(id);
     _audioHandler.inner = _playerAudioHandler;
     _audioHandler.playbackState.listen((playbackState) {
       broadcastPlaybackEvent();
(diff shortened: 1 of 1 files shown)


Zekfad:
@ryanheise here's discussed prototype.
Default behavior is preserved and working (I've done testing on chrome and my android phone).
I've also added `PlayerStateSnapshot` to expose state, because `_PlayerAudioHandler` is internal, has many methods and properties that are undesirable to be exposed.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -165,7 +171,7 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
   late final _PlayerAudioHandler _playerAudioHandler;
 
   _JustAudioPlayer({required String id}) : super(id) {
-    _playerAudioHandler = _PlayerAudioHandler(id);
+    _playerAudioHandler = _PlayerAudioHandler._(id);
```
ryanheise: This change is unnecessary since the class itself is private.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -839,3 +859,42 @@ extension _ValueStreamExtension<T> on ValueStream<T> {
...
+  /// More info: https://developer.android.com/about/versions/13/behavior-changes-13#playback-controls
+  final List<int>? androidCompactActionIndices;
+  final List<MediaControl> controls;
+  final Set<MediaAction> systemActions;
+}
+
+
+/// Snapshot of current player state.
+/// Used to provide data to [NotificationConfigBuilder].
+abstract class PlayerStateSnapshot {
```
ryanheise: It's unfortunate to replicate existing state classes that are in just_audio, but just_audio_background should not depend on just_audio so this is understandable for now.

Eventually, I think these state classes should be defined in just_audio_platform_interface where they can be reused.

just_audio_background/lib/just_audio_background.dart
```diff
@@ -839,3 +859,42 @@ extension _ValueStreamExtension<T> on ValueStream<T> {
...
+  /// -1 if no next index available
+  int get nextIndex;
+  /// -1 if no previous index available
+  int get previousIndex;
+  /// Whether player is currently in playing state.
+  bool get playing;
+  /// Whether player is currently paused.
+  bool get paused => !playing;
```
ryanheise: I'd rather not have a redundant `paused` since it has no basis in just_audio's existing API design.

just_audio_background/lib/just_audio_background.dart
Zekfad: Ah, yes, that remained from my first attempt, will re vert this.

just_audio_background/lib/just_audio_background.dart
Zekfad: So, is this in scope of this PR, if so, tell me where I should move it, please?

just_audio_background/lib/just_audio_background.dart
Zekfad: I thought it would be convenient for end user to have paused getter, bit I'll remove it then, no prob

just_audio_background/lib/just_audio_background.dart
ryanheise: That comment was just to recognise the need for something now even though it should be moved elsewhere in the future.

But as it is, is  this really a "snapshot"? It looks like you're using this as a mixin (although you defined it as an abstract class instead of a mixin). That means that this package is now returning a reference to an internal class which it probably shouldn't. If the so-called snapshot is actually an object where the values may change behind the scenes, then it is not really a snapshot. I think it would be far less surprising to create an immutable state object, and to build a new instance of it each time a new state needs to be broadcast.

As for naming, and considering where these state objects will likely move in the future, I would say we do not know at this stage what that future state object will look like, and anything we might come up with now would be rather narrow in its vision. So to avoid any future problems, it might be best to choose a class name for this state that is obviously specific to this particular package rather than some name that we might want to reserve for the future ideal API.

ryanheise:
Just as an update on where I'm at with this PR, I am not yet 100% sure on which way we want to go on the API design.

If we look at the default implementation of the notification builder:

```dart
_defaultNotificationConfigBuilder(PlayerStateSnapshot state) {
    final controls = [
      if (state.hasPrevious) MediaControl.skipToPrevious,
      if (state.playing) MediaControl.pause else MediaControl.play,
      MediaControl.stop,
      if (state.hasNext) MediaControl.skipToNext,
    ];
    return NotificationConfig(
        controls: controls,
        systemActions: const {
          MediaAction.seek,
          MediaAction.seekForward,
          MediaAction.seekBackward,
        },
        androidCompactActionIndices: List.generate(controls.length, (i) => i)
          ..removeAt(controls.indexOf(MediaControl.stop)));
  }
```

I still find this `PlayerStateSnapshot` problematic. Of course as discussed I think eventually just_audio should be restructured so that the state classes are moved into the platform interface package where they can be reused, so we don't have this redundant data structure. Right now, it is not possible to, say, declare the parameter type of this builder function to be `AudioPlayer` since just_audio_background cannot introduce a compile time dependency on just_audio. However, the app itself CAN have a compile time dependency on both just_audio and just_audio_background, and so one option is to simply remove this parameter, and allow the app to hold onto its own reference to the `AudioPlayer` instance and query its state.

In terms of API naming, an app may display other kinds of notifications, and so it may perhaps be a good idea to prefix these class names with `Media`. So, `MediaNotificationConfig` and `MediaNotificationConfigBuilder` (and also see if it's possible to delete `PlayerStateSnapshot`).

PR #1043, opened 2023-08-18
Stop passing a nullable value to Completer<nn-type>.completer.

srawlins:
This is cleanup work required to start enforcing this with static analysis, as per https://github.com/dart-lang/sdk/issues/53253.

Real quick this issue is that this code is unsafe:

```dart
void f(Completer<int> c, int? i) {
  Future<int>.value(i); // Ouch!
  c.complete(i);        // Ouch!
}
```

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3771,9 +3771,9 @@ class AndroidEqualizer extends AudioEffect with AndroidAudioEffect {
     }
     final response = await platform
         .androidEqualizerGetParameters(AndroidEqualizerGetParametersRequest());
-    final parameters =
+    final receivedParameters =
         AndroidEqualizerParameters._fromMessage(_player!, response.parameters);
-    _parametersCompleter.complete(parameters);
+    _parametersCompleter.complete(receivedParameters);
   }
 
   /// The parameter values of this equalizer.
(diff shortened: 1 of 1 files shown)


just_audio/lib/just_audio.dart
```diff
@@ -3772,9 +3771,9 @@ class AndroidEqualizer extends AudioEffect with AndroidAudioEffect {
     }
     final response = await platform
         .androidEqualizerGetParameters(AndroidEqualizerGetParametersRequest());
-    _parameters =
+    final parameters =
```
ryanheise: This local variable shadows the getter of the same name and causes an error on line 3769: "Local variable 'parameters' can't be referenced before it is declared".

You might either rename the local variable to something like `receivedParameters` or change line 3769 to use `this.parameters`.

ryanheise:
Thanks for the heads up and PR, although I note there is a small compile error in the PR(see comment).

If no further commits are pushed, I'd be happy to make the change on my end and merge tomorrow.

just_audio/lib/just_audio.dart
srawlins: Yikes sorry for the noise. Thanks for the suggestion. Applied.

ryanheise:
Thanks @srawlins :+1: . Merged.

Issue #966, opened 2023-03-31
Prepare for uuid 4.0.0


ryanheise:
**Is your feature request related to a problem? Please describe.**

just_audio currently depends on uuid 3.x. Once uuid 4 is released, people won't be able to use it with just_audio until we relax our constraints to support it. 

**Describe the solution you'd like**

Update the pubspec constraint to allow both uuid 3 and uuid 4. (The part of uuid we're using is compatible with both major versions.)

**Describe alternatives you've considered**

Remove the dependency on uuid. Although that's not necessary here.

**Additional context**

devyueyue:
Hi，How to play music in the background and compatible with the playback completion status. When the audio playback app is switched to the background, the playback progress stops refreshing

devyueyue:
Hi!   
How to monitor the playback completion status when playing music in the background (it is necessary to monitor the playback completion status and upload data to the server). When the audio playback app is switched to the background, the playback progress stops refreshing

devyueyue:
![image](https://user-images.githubusercontent.com/96807013/229408538-04cfb2b4-90da-4637-b234-b9db0a4a6dbd.png)

When the app switches to the background, the stream stops refreshing

phamnhuvu-dev:
@ryanheise ryanheise
uuid v4 just fixed a critical issue, and we really want to update that version
https://github.com/daegalus/dart-uuid/issues/87

ryanheise:
You can try forcing that with a dependency override, but since version 4 of the package hasn't been released as stable yet, I will wait until it is officially released as stable.

mikhael28:
@ryanheise hey there, good morning! I wanted to let you know that the uuid v4 has now been released as stable as of 5 days ago. It might be a good time to take a look, and see if now the time is right to update to fix the dependency constraints issue.

On a separate note, I wanted to let you know that your package is excellent and perfectly provides the functionality we were looking for in a simple and straightforward way. If I may be so bold, when you integrate the uuid changes, I would recommend bumping up the version number of the package to 1.0 from 0.9.x

A library of such quality deserves to have a non-zero digit to the far left :)

ryanheise:
Hi @mikhael28 , I am aware but I'm unfortunately in the process of upgrading my system and I don't have the Internet fully restored. But it should happen this weekend. @Pante is also in the process of contributing a PR (#1054 ) .

> A library of such quality deserves to have a non-zero digit to the far left :)

[...]

PR #1054, opened 2023-09-08
Upgrade UUID dependency from ^3.0.1 to ^4.0.0

Pante:
This is a small PR that upgrades the minimum UUID dependency from ^3.0.1 to ^4.0.0 which supports the [new UUID formats](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-uuidv7-layout-and-bit-order). I'm not really sure if this constitutes as a breaking change. Please do let me know if this should be branched off the `major` branch instead.

### DIFF
--- a/just_audio/pubspec.yaml
+++ b/just_audio/pubspec.yaml
@@ -25,7 +25,7 @@ dependencies:
   path: ^1.8.0
   path_provider: ^2.0.0
   async: ^2.5.0
-  uuid: ^3.0.1
+  uuid: ^4.0.0
   crypto: ^3.0.0
   meta: ^1.3.0
   flutter:


ryanheise:
Hi @Pante yes that would be a breaking change. Please note #966 :

> Update the pubspec constraint to allow both uuid 3 and uuid 4. (The part of uuid we're using is compatible with both major versions.)

Pante:
My bad, seems like I missed the issue. I'll reopen another PR that addresses both issues.

EDIT: Actually would it still be a breaking change if the constraint was changed to `'>=3.0.1 <5.0.0'` instead?

ryanheise:
It's not a breaking change if it still supports the versions currently supported. So that's fine. You can also see how I've done it for `rxdart`.

PR #1062, opened 2023-09-17
Set the right Playback state to pass Android Media Controller Tests.

iosifnicolae2:
For complete information please take a look at: https://github.com/ryanheise/audio_service/pull/1042#issuecomment-1722520668


iosifnicolae2:
If @ryanheise consider this PR useful, it can be opened again and merged, until then I'll close it.

Thank you!

ryanheise:
This would not be a correct solution to your problem, as just_audio's states are not intended to be based on the legacy media session states, they are based on ExoPlayer states. In particular, `playing` is not a processing state, it is orthogonal to the processing state. So you can having playing+ready or playing+buffering or !playing+ready or !playing+buffering.

In the same fashion as ExoPlayer, just_audio's states are "mapped" into the legacy media session states. I was trying to work with you on the other issue to come up with a proper solution.

PR #1072, opened 2023-09-24
FIX: Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type cast

goviral-ma:
Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type cast

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -864,7 +864,7 @@ class AudioPlayer {
   /// original [AudioSource]. If [end] is null, it will be reset to the end of
   /// the original [AudioSource]. This method cannot be called from the
   /// [ProcessingState.idle] state.
-  Future<Duration?> setClip({Duration? start, Duration? end}) async {
+  Future<Duration?> setClip({Duration? start, Duration? end, dynamic tag: tag}) async {
     if (_disposed) return null;
     _setPlatformActive(true)?.catchError((dynamic e) async => null);
     final duration = await _load(
@@ -875,6 +875,7 @@ class AudioPlayer {
                 child: _audioSource as UriAudioSource,
                 start: start,
                 end: end,
+                tag: tag
               ));
     return duration;
   }
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -864,7 +864,7 @@ class AudioPlayer {
   /// original [AudioSource]. If [end] is null, it will be reset to the end of
   /// the original [AudioSource]. This method cannot be called from the
   /// [ProcessingState.idle] state.
-  Future<Duration?> setClip({Duration? start, Duration? end, dynamic tag: tag}) async {
+  Future<Duration?> setClip({Duration? start, Duration? end, dynamic tag}) async {
     if (_disposed) return null;
     _setPlatformActive(true)?.catchError((dynamic e) async => null);
     final duration = await _load(


otabekoff:
I think I am also getting this error when trying to use `just_audio_cache` with `just_audio` and `just_audio_background` packages together. I have predefined data and when I want to set them into a MediaItem it is giving "[Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type cast](https://github.com/ryanheise/just_audio/pull/1072#top) error. I thought it was happening by me. But it was the error gotten by others too. Please, fix this problem. As for now I am using type casting like `state........data.get as MediaItem(.....` as a workaround.

otabekoff:
I my opinion, these changes by @goviral-ma doesn't fix errors at all.  Any further fixes @ryanheise ???

otabekoff:
@goviral-ma I tested the reason why it is happening. When I removed `dynamicSet` part from `init`, it began to work. I mean it is issue with `just_audio_cache` plugin which is not owned by Ryan Heise.

goviral-ma:
@otabekoff in my case I'm not using `just_audio_cache`
I'm using `just_audio` with `just_audio_background`, and this is a quick fix that worked for me and solved my problem.
The function `setClip` in just_audio doesn't take tag as parameter, and in the core of the function, an instance of `ClippingAudioSource` is created with null as tag variable, that's why the problem produced.

I simply solved it by adding tag variable to `setClip` to pass the value of tag to the function.

otabekoff:
@goviral-ma So caching is working for you?

mathisfouques:
Hey @otabekoff , in this [file](https://github.com/lau1944/just_audio_cache/blob/main/lib/src/audio_player_utils.dart), dynamicSet method use setFilePath from just_audio package. In this method [setFilePath](https://github.com/ryanheise/just_audio/blob/6952be55c6ddfb01f2567c0737ef098d1bd9fd45/just_audio/lib/just_audio.dart#L701), you see that AudioSource.file() does not set the tag, which is also added in this PR (but for setClip method). 

I've come up with this message because I had the same problem, but using just_audio + just_audio_background, with the setFilePath method, which, as setClip, does not set the tag used in just_audio_background. 

If you're just using just_audio + [...]_background with setFilePath, consider manually setting a tag as MediaItem to solve this : 
```
await player.setFilePath(path);

await player.setAudioSource(AudioSource.file(path,
  tag: MediaItem(
   id: path,
   title: path.split('/').last,
)));
```

Anyway, I think this notion of tag is not clear. I'm not sure if the value of the tag can ever NOT be a MediaItem, and if not we should consider removing the dynamic typing by MediaItem? . And even remove the nullable and put a default value ?
@ryanheise thoughts ? I can put another PR if you want.

ryanheise:
First, thanks for bumping this, since it went off my radar. The tag is for associating your app's own data with each audio source. It's similar to Android's [View tags](https://developer.android.com/reference/android/view/View#tags). Any data structure can be stored in it. If you use just_audio_background, then you are letting that plugin manage the tag for you, so you don't get a choice, it has to be MediaItem. But if you want to manage things yourself, you should not use just_audio_background (i.e. use audio_service directly), and then you can use the tag to store whatever data you find useful.

Regarding the PR, yes this is important, and I agree the tag parameter probably needs to be added to a number of other places as well.

mathisfouques:
Ok, here is the PR #1219 (@ryanheise)

ryanheise:
Thanks, @goviral-ma for the contribution. This is now merged into `minor` along with #1219 for the other missing ones.

Issue #1032, opened 2023-07-26
LockCachingAudioSource._fetch  _File.renameSync issue


AmmarZY:
The following error is reported by firebase analytics regarding LockCachingAudioSource._fetch method, where an exception is thrown by _File.renameSync (dart:io)

**Error messages**

Fatal Exception: FlutterError
PathNotFoundException: Cannot rename file to '/var/mobile/Containers/Data/Application/DFBCF9C9-13D9-4E3D-853D-7487BF63D8A2/Library/Caches/todayprayers_audio/zeyarat_ashuraa_abather.mp3', path = '/var/mobile/Containers/Data/Application/DFBCF9C9-13D9-4E3D-853D-7487BF63D8A2/Library/Caches/todayprayers_audio/zeyarat_ashuraa_abather.mp3.part' (OS Error: No such file or directory, errno = 2)

Fatal Exception: FlutterError
0  ???                            0x0 _File.renameSync (dart:io)
1  ???         

No errors occur

- Iphone 14
- version: 16.4.1
Futter SDK 3.10.4

PR #1076, opened 2023-09-30
Fix PathNotFoundException when using LockCachingAudioSource

Colton127:
Relevant issue: https://github.com/ryanheise/just_audio/issues/1032

I have this issue littered throughout my sentry logs. I can sometimes replicate it by setting a LockCachingAudioSource and then disposing of the audio player almost immediately afterwards. It appears that just_audio attempts to rename (or delete on error) the partial cache file, but does so without first checking to see if it exists. This pull request simply ensures the partial cache file exists before attempting to rename or delete it.


Colton127:
@ryanheise Any updates on this? Thank you for your time.

danielgomezriconeon:
@ryanheise ping

ryanheise:
Thank you for the ping - I've been busy on the next major release for the past few weeks, but should look at this one after that is released. Please feel free to ping me again in a week if I haven't attended to this by then.

PR #1092, opened 2023-10-21
Fix conversion between milliBel and deciBel

### LINKED ISSUE #1089: Gain Factor for AndroidLoudnessEnhancer is wrong
https://pub.dev/documentation/just_audio/latest/just_audio/AndroidLoudnessEnhancer/setTargetGain.html
Docs state:  
> Sets the target gain to a value in **decibels**.

However, the value actually seems to be in 10 decibels.  
See implementation here:
https://github.com/ryanheise/just_audio/blob/3622d11f14d4191e4c1326c32c7a025941af40ec/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java#L763-L766

The used factor is `1000`, whereas the Android documentation ([link](https://developer.android.com/reference/android/media/audiofx/LoudnessEnhancer)) states:  
> The maximum gain applied applied to the signal to process. It is expressed in **millibels (100mB = 1dB)** where 0mB corresponds to no amplification.

So the correct factor appears to be `100`

The example

1. Apply and activate the AndroidAudioEnhancer effect
2. Lower the volume so that the playing track is fairly quiet (make sure it is *really low*! Otherwise you might damage your speakers...)
3. Now set the gain to 10, which should be equal to 10 dB and therefore seem about twice as loud
4. Get your eardrums blows out, since the audio is definitely more than twice as loud :D

**Error messages**

```
n/a
```

**Expected behavior**

The docs say the gain value is in decibels, so setting gain to `10.0` should result in 10 dB gain.

**Screenshots**

 - OS: n/a
 - Browser [e.g. chrome, safari + version]

 - Device: Asus Zenfone 10
 - OS: Android 13

**Flutter SDK version**
```
[√] Flutter (Channel stable, 3.13.5, on Microsoft Windows [Version 10.0.19045.3570], locale en-US)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[√] Android toolchain - develop for Android devices (Android SDK version 34.0.0-rc4)
[√] Chrome - develop for the web
[√] Visual Studio - develop Windows apps (Visual Studio Build Tools 2019 16.11.1)   
[√] Android Studio (version 2022.2)
[√] Android Studio (version 2022.1)
[√] VS Code (version 1.83.1)       
[√] Connected device (4 available) 
[√] Network resources
```

**Additional context**

I acknowledge that the unit "Millibel" is a bit confusing in this case, but this is because the base unit is **deci**bel, so 10 bel. The math does work out ^^

Chaphasilor:
As discussed in #1089, here is the PR that fixes the issue.  
I went ahead and also applied the fix to other methods that had the same problem (e.g. equalizer).

This closes #1089

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.10.1
 
-* 
+* Fix conversion between milliBel and deciBel (@Chaphasilor).
 
 ## 0.9.35
 
--- a/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -666,7 +666,7 @@ public class AudioPlayer implements MethodCallHandler, Player.Listener, Metadata
         case "AndroidLoudnessEnhancer":
             if (Build.VERSION.SDK_INT < 19)
                 throw new RuntimeException("AndroidLoudnessEnhancer requires minSdkVersion >= 19");
-            int targetGain = (int)Math.round((((Double)map.get("targetGain")) * 1000.0));
+            int targetGain = (int)Math.round((((Double)map.get("targetGain")) * 100.0)); // target gain needs to be provided in milliBel, the user provides the value in deciBel
             LoudnessEnhancer loudnessEnhancer = new LoudnessEnhancer(audioSessionId);
             loudnessEnhancer.setTargetGain(targetGain);
             return loudnessEnhancer;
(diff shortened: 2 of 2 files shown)


Chaphasilor:
@ryanheise is there anything that you need me to add to this? Or can we get this merged? :)

ryanheise:
Thank you for the reminder. I am currently in the middle of a simulation which is taking up all of my RAM for the moment (for a few days) but I've got this on my list as the first PR to merge once it's done.

ryanheise:
Simulation finished, will work on this when I get home this afternoon.

ryanheise:
Tested and merged into `major`. Thanks again for finding and fixing this.

Would it cause you any inconvenience if took the opportunity to work on some other breaking changes to go along with this, or do you depend on having a new major release soon?

PR #1102, opened 2023-10-24
Fix headers passed as null with uri source on iOS

fmaerkl-sw:
`data[@"headers"]` is not nil, but `NSNull` by default, which will cause an exception in the `initWithId: ...` call

### DIFF
--- a/just_audio/darwin/Classes/AudioPlayer.m
+++ b/just_audio/darwin/Classes/AudioPlayer.m
@@ -452,12 +452,13 @@
 
 - (AudioSource *)decodeAudioSource:(NSDictionary *)data {
     NSString *type = data[@"type"];
-    if ([@"progressive" isEqualToString:type]) {
-        return [[UriAudioSource alloc] initWithId:data[@"id"] uri:data[@"uri"] loadControl:_loadControl headers:data[@"headers"]];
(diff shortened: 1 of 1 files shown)


ryanheise:
Thanks for noticing the bug. I also ended up fixing it myself which I haven't pushed yet, although I handled it inside `UriAudioSource` without touching `decodeAudioSource` which I want to leave as is to match the same cases used in other platform implementations.

I don't mind giving you credit nonetheless in the changelog, or if you modify your PR in the same way I could merge it.

fmaerkl-sw:
Feel free to push your own fix, it will probably be faster than modifying the pr.

ryanheise:
I've pushed my fix to `feature/extractor_options`. (Sorry it was not faster, but dealing with hand pain).

If you don't mind testing it, I would be interested to know if you notice any other issues with it. If all goes well, I'll publish these changes to the platform interface so that we can avoid having to make the changes to the `pubspec.yaml`.

fmaerkl-sw:
@ryanheise Thanks. Just tested again and everything seems to work.

ryanheise:
Thanks for the quick response. I'll try to finish up the changes to the platform interface tomorrow morning and then publish.

ryanheise:
Now published.

PR #1104, opened 2023-10-26
Fix unable to overwrite request headers

ctedgar:


ryanheise:
Thanks for the PR and especially for the test case. Since I have some queued changes for 0.9.35 waiting on the `feature/extractor_options` branch, I will consider merging this into that branch.

ryanheise:
This has now been merged into `minor` after merging all of the other aforementioned changes. Would you mind having a quick test to check if your use case works?

Also note that one of the other changes made was to support the platform's native implementation of headers instead of the proxy with the caveat that on iOS this makes use of an undocumented API which everybody uses but in theory gives Apple the right to perhaps reject your app under the app store review. By default, it will use the proxy as before, but there is a constructor parameter to switch between the two behaviours.

ryanheise:
I've gone ahead and published the next release.

ctedgar:
> This has now been merged into `minor` after merging all of the other aforementioned changes. Would you mind having a quick test to check if your use case works?
> 
> Also note that one of the other changes made was to support the platform's native implementation of headers instead of the proxy with the caveat that on iOS this makes use of an undocumented API which everybody uses but in theory gives Apple the right to perhaps reject your app under the app store review. By default, it will use the proxy as before, but there is a constructor parameter to switch between the two behaviours.

Hi @ryanheise, 

I did a test w/ 0.9.36 and all is working as expected.  I'll close the original issue. 
Thanks for the heads up about the native impl of headers; in my case I defer the option of making http connections to the user so I'm happy with the proxy.

Thanks again,

Cris

Issue #1121, opened 2023-11-13
Browser cannot capture exceptions thrown by .play() method


idy:
When I call audioPlayer.play() method, the method throws a NotAllowedError exception. I tried to capture the exception using try-catch, but the exception is not being caught.

PR #1122, opened 2023-11-13
Fix: NotAllowedError can be captured by try-catch

idy:
Relevant issue: #1121

### DIFF
--- a/just_audio_web/lib/just_audio_web.dart
+++ b/just_audio_web/lib/just_audio_web.dart
@@ -961,10 +961,15 @@ class _PlayPauseQueue {
 
   Future<void> _run() async {
     await for (var request in _queue.stream) {
-      if (request.playing) {
-        await audioElement.play();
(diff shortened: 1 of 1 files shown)


just_audio_web/lib/just_audio_web.dart
```diff
@@ -961,10 +961,15 @@ class _PlayPauseQueue {
...
-        audioElement.pause();
+      try {
+        if (request.playing) {
+          await audioElement.play();
+        } else {
+          audioElement.pause();
+        }
+      } catch (err) {
+        request.completer.completeError(err);
+        continue;
```
ryanheise, line 972: Rather than use `continue`, this can be handled by the try/catch control ostructure by including the normal `complete()` from below within the `try` block. I can fix this while merging.

ryanheise:
Thank you very much for the PR, it is now merged and published.

Issue #1030, opened 2023-07-22
 Error using Chinese path


fangjunai:
**Which API doesn't behave as documented, and how does it misbehave?**
Name here the specific methods or fields that are not behaving as documented, and explain clearly what is happening.

**Minimal reproduction project**
```
import 'package:flutter/material.dart';
import 'package:just_audio/just_audio.dart';

class Home extends StatefulWidget {
  const Home({Key? key}) : super(key: key);

  @override
  State<Home> createState() => _HomeState();
}

class _HomeState extends State<Home> {
  final player = AudioPlayer();
  String path = "C:\\Users\\FangJun\\Music\\123.mp3";  //normal
  String path2 = "C:\\Users\\FangJun\\Music\\莫问归期 - 蒋雪儿.mp3"; //Error using Chinese path

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(title: const Text('Music')),
      body: Center(
        child: Column(
          children: [playButton(), stopButton()],
        ),
      ),
    );
  }

  Widget playButton() {
    return ElevatedButton(
      onPressed: doPlay,
      child: const Text("Play"),
    );
  }

  Widget stopButton() {
    return ElevatedButton(
      onPressed: doStop,
      child: const Text("Stop"),
    );
  }

  void doPlay() async {
    await player.setAudioSource(AudioSource.file(path2)); //Error using Chinese path
    await player.play();
  }

  void doStop() async {
    await player.stop();
  }
}
```

**To Reproduce (i.e. user steps, not code)**

**Error messages**
```
[just_audio_windows] Called setVolume
[just_audio_windows] Called setSpeed
[just_audio_windows] Called setPitch
[just_audio_windows] Called setSkipSilence
[just_audio_windows] Called setLoopMode
[just_audio_windows] Called setShuffleMode
[just_audio_windows] Called load
[just_audio_windows] Media error:
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Loading interrupted
#0      AudioPlayer._load
<asynchronous suspension>
#1      AudioPlayer._setPlatformActive.setPlatform
<asynchronous suspension>
```

**Expected behavior**
Normal playback of music with Chinese path

**Screenshots**

**Desktop (please complete the following information):**
 - OS: win10 x64
 - 
**Smartphone (please complete the following information):**

**Flutter SDK version**
```
PS C:\Users\FangJun> flutter doctor
[...]

PR #1209, opened 2024-03-21
Fix: Error using Chinese path

YunHerry:
fix issue: (#1030)

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -2375,7 +2375,7 @@ class ProgressiveAudioSource extends UriAudioSource {
   @override
   AudioSourceMessage _toMessage() => ProgressiveAudioSourceMessage(
         id: _id,
-        uri: _effectiveUri.toString(),
+        uri: Uri.decodeFull(_effectiveUri.toString()),
         headers: _mergedHeaders,
         tag: tag,
         options: options?._toMessage(),
@@ -2404,7 +2404,7 @@ class DashAudioSource extends UriAudioSource {
   @override
   AudioSourceMessage _toMessage() => DashAudioSourceMessage(
         id: _id,
-        uri: _effectiveUri.toString(),
+        uri: Uri.decodeFull(_effectiveUri.toString()),
         headers: _mergedHeaders,
         tag: tag,
       );
@@ -2431,7 +2431,7 @@ class HlsAudioSource extends UriAudioSource {
   @override
   AudioSourceMessage _toMessage() => HlsAudioSourceMessage(
         id: _id,
-        uri: _effectiveUri.toString(),
+        uri: Uri.decodeFull(_effectiveUri.toString()),
         headers: _mergedHeaders,
         tag: tag,
       );


ryanheise:
I don't believe this is the correct solution. For Windows, I think the `Uri` should be created with `Uri.file(s, windows: true)`.

YunHerry:
in debugging, the value of `_uri.toString()` in `_toMessage` method is not a correct value. although I used `Uri.file(s, windows: true)`, but the value of uri is stiil incorrect in debug. the value I get from `_uri.toString()` is file:///E:/music_cache/%E4%B8%AD.ogg, but, what I want is file:///E:/music_cache/中.ogg
d1593eb0c3ba0c2a834edb0b3afe382f

ryanheise:
Why would you want it to be `file:///E:/music_cache/中.ogg`? That is not a valid URI according to the standard, so it should not be done.

YunHerry:
it may not comply with the standard, but it's difficult for us to predict the file names of the user's environment

ryanheise:
This is not about predicting file names, `中.ogg` is a valid and correct filename. The solution is simply to convert a valid URI into a valid filename on the platform side.

YunHerry:
how should we solve this conversion problem

ryanheise:
When I say the "platform side", I am referring to the native implementation for the platform. You can see how I have implemented it for iOS  (one objective C) and Android (in Java), but the windows implementation is not provided by this repository, it is provided by whichever additional plugin you are using. Which plugin are you using to provide the windows implementation? That is where you will find the platform side of the implementation where it takes the valid Uri and tries to convert that into whatever the platform requires.

If the Uri is not valid in the first place, then we need to fix that first here in just_audio, but then the conversion to the platform side happens in the other plugin that provides the windows platform implementation.

YunHerry:
All plugins communicate based on `_ToMessage()`, why not transcode it in this library, but convert it in the plugin?

ryanheise:
I'm sorry I don't understand what you're referring to by the distinct terms plugin and library which seem the same. However, I can't introduce breaking changes to the platform interface, a Uri is the correct platform independent representation of a resource, and it is still up to the specific platform implementation to handle it in it's own way.

ryanheise:
I assume you are using [just_audio_windows](https://github.com/bdlukaa/just_audio_windows/tree/master/just_audio_windows) to implement the platform side? If so, digging into the [source code](https://github.com/bdlukaa/just_audio_windows/blob/master/just_audio_windows/windows/player.hpp) I see it takes the URI you give it and handles it on the platform in this way:

```cpp
  /**
  * Creates a single MediaSource.
  */
  MediaSource AudioPlayer::createMediaSource(const flutter::EncodableMap& source) const& {
    const std::string* type = std::get_if<std::string>(ValueOrNull(source, "type"));
    if (type->compare("progressive") == 0 || type->compare("dash") == 0 || type->compare("hls") == 0) {
      const auto* uri = std::get_if<std::string>(ValueOrNull(source, "uri"));
      return MediaSource::CreateFromUri(
        Uri(TO_WIDESTRING(*uri))
      );
    } else {
      throw std::invalid_argument("Source is unsupported or can not be nested: " + *type);
    }
  }
```

So it is letting the Windows API directly handle the URI. From the documentation I have seen, Chinese characters should be URL-encoded. I believe that is true, but if it is not, any further rewriting of the URI to work around Window's non-standard handling of URIs will need do be done in the just_audio_windows package rather than this one.

What I can do over here in just_audio is possibly add a `bool windows` parameter to `setFilePath`, although in the meantime, users can instead construct their own Uri via `setAudioSource`.

I would be interested to learn of any progress you make.

canxin121:
> All plugins communicate based on `_ToMessage()`, why not transcode it in this library, but convert it in the plugin?

我认为实际上这应该是WinRT的bug，而不是dart的Uri的问题，估计很快可以打个补丁修上

ryanheise:
@canxin121 do you have a preference on whether the `windows: true` parameter on the Dart side is preferable?

Issue #991, opened 2023-05-12
Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type 


chitraarasu:
**Which API doesn't behave as documented, and how does it misbehave?**
Name here the specific methods or fields that are not behaving as documented, and explain clearly what is happening.

**Minimal reproduction project**
Provide a link here using one of two options:

**To Reproduce (i.e. user steps, not code)**

**Error messages**

```
Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type 
```

**Expected behavior**
I am facing issue white try to play. yesterday tried to integrate this package that time worked fine. unfortunately deleted code due to some reason. Now again trying to integrate on same way but getting this issue again and again. trying to fix about 2 days pls help me to fix this issue

**Screenshots**
<img width="1229" alt="image" src="https://github.com/ryanheise/just_audio/assets/75994428/d3545d01-9ff1-4b13-a150-d6179d5dab2d">

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: iPhone 14 pro
 - OS: 

**Flutter SDK version**

**Additional context**

ryanheise:
Closing this issue (as promised in the message in your last issue).

chitraarasu:
What is solution for this issue. I have followed documentation properly but still getting this error.

ryanheise:
I guess you're just going to continue ignoring me, and I don't have time for that. Locking the issue. (If you want more context, please read the comment on your previous issue which you ignored. Then please read the instructions for posting an issue which you also ignored.)

PR #1219, opened 2024-04-02
FIX : tag is null when using setFilePath/setAsset/setUrl along with just_audio_background. 

mathisfouques:
Solves the following error message when using just_audio_background : "Unhandled Exception: type 'Null' is not a subtype of type 'MediaItem' in type cast". It's likely to solve #991 as well.  

just_audio_background needs it as described [here](https://github.com/ryanheise/just_audio/pull/1072#issuecomment-2031815910). 

Changes on just_audio and the background package. Adding tags on convenience functions, updating doc, and putting an assert with correct debug message before the cast to avoid getting the same error message in the future.

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -673,7 +673,7 @@ class AudioPlayer {
   /// This is equivalent to:
   ///
   /// ```
-  /// setAudioSource(AudioSource.uri(Uri.parse(url), headers: headers),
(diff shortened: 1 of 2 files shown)


rahulnpadalkar:
Is this going to be merged in? If not is there any workaround for now?

ryanheise:
Thanks @mathisfouques for the contribution. This has now been merged into `minor` with some "minor" changes to the comments/assertions. Let me know if I've broken anything for you.

PR #1220, opened 2024-04-04
iOS - Added Privacy Manifests

martin-headspace:
# Summary
Added an empty Privacy Manifest to the iOS project in preparation for the May 1st Deadline for 3rd Party SDKs that include native iOS code. 

# Description

> Starting May 1: You’ll need to include approved reasons for the listed APIs used by your app’s code to upload a new or updated app to App Store Connect.
> ...
> Make sure to use a version of the SDK that includes its privacy manifest and note that signatures are also required when the SDK is added as a binary dependency.

[Privacy updates for App Store submissions - Latest News - Apple Developer](https://developer.apple.com/news/?id=3d8a9yyh)
[Upcoming third-party SDK requirements - Support - Apple Developer](https://developer.apple.com/support/third-party-SDK-requirements/)

# Impact and Testing
Developers using an updated version of this library on their iOS projects shouldn't see any issues related to the Privacy Manifest

# Contributor Note
Based on the language on Apple's Website, this change seems to be mandatory for all 3rd Party Libraries. The impact of not adding it is unclear to me, but hopefully having it will be better than not.

### DIFF
--- a/just_audio/ios/Resources/PrivacyInfo.xcprivacy
+++ b/just_audio/ios/Resources/PrivacyInfo.xcprivacy
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyTrackingDomains</key>
+	<array/>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array/>
+	<key>NSPrivacyCollectedDataTypes</key>
+	<array/>
+	<key>NSPrivacyTracking</key>
+	<false/>
+</dict>
+</plist>
--- a/just_audio/ios/just_audio.podspec
+++ b/just_audio/ios/just_audio.podspec
@@ -17,5 +17,6 @@ A new flutter plugin project.
   s.dependency 'Flutter'
   s.platform = :ios, '8.0'
   s.pod_target_xcconfig = { 'DEFINES_MODULE' => 'YES', 'VALID_ARCHS[sdk=iphonesimulator*]' => 'x86_64' }
+  s.resource_bundles = {'just_audio' => ['Resources/PrivacyInfo.xcprivacy']}
 end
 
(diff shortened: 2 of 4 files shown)


himanshu-newstok:
There are only 5 days left. @ryanheise   What are your plans for adding this in all your packages like `just_audio`,`just_audio_background`.`audio_session`,`audio_service`

ryanheise:
My contact from Apple for this matter said he is trying to get an answer for me before the deadline but as yet that hasn't materialised. My own understanding is that a privacy manifest is not required for a plugin that doesn't use any of the APIs in question, but we will see. I will either merge this PR or not depending on which way proves to be correct in 5 days. Therefore, if you are in the situation in 5 days of an app store rejection that can be traced specifically to this plugin, please try this PR and if it works, I will merge and publish a new release.

ryanheise:
One more thing to mention that "if" it turns out that privacy manifests are required, I would assume they would need to be added to both this plugin AND the audio_session plugin, because audio_session contains code (albeit code that can be compiled out) to access sensitive devices like the microphone. However, checking the list of APIs that require a privacy manifest, these microphone APIs simply aren't on the list. But we'll need to react to whatever the case turns out to be in a few days (first person to request App Store approval, please let us know your experiences.)

himanshu-newstok:
The warning I am currently getting are `NSPrivacyAccessedAPICategoryDiskSpace` , `NSPrivacyAccessedAPICategoryUserDefaults` and `NSPrivacyAccessedAPICategoryFileTimestamp` in Runner file. Does this plugin uses any of these apis. If not then the plugin do not need to include anything else. I also have many other packages that can cause these warning, so I am not sure about if these warning is due to your plugin or not.

ryanheise:
No, this plugin doesn't use any of those. `NSPrivacyAccessedAPICategoryUserDefaults` is probably caused by shared_preferences which I don't use. Now, audio_service does use the SharedPreferences API directly on Android, but its counterpart is not used on iOS.

Abeansits:
@ryanheise since the Apple requirement is rolling out tomorrow, how do you feel about merging this?

ryanheise:
@Abeansits Please read the comments above.

Abeansits:
Didn't read them properly thank you @ryanheise. So the plan is to wait and see what Apple actually cares about, love it. :)

Issue #870, opened 2022-11-08
There seems to be no way when using justaudio_background to catch errors -> All Unhandled


matecode:
First of all thanks for solving #832, i tried it and everything seems to work as expected so far. As i cannot write in the old bug report anymore (yes github, i was off for holiday for 14 days :) ) I use a new one.

But now I'm facing a new issue:

**Which API doesn't behave as documented, and how does it misbehave?**
I cannot catch the errors thrown by the async methods of an AudioPlayer instance if justaudio_background is used, namely load() and play(). The errors are uncaught even if i surround the method calls. I tried with AudioSource.uri and LockCachingAudioSource ... i first thought the problem is LockCachingAudioSource (thats why i named the demo branch) but it seems it is more the justaudio_background handler.

Also playbackEventStream does not give an Error (should it!?)

**Minimal reproduction project**
https://github.com/matecode/just_audio/tree/Lock_Caching_Error_Handling_Bug

To demonstrate I changed the urls of audioSources to not existent links.

As you can see in main.dart in line
**72**  -> setAudioSource should be catched (preload must be true) because load() is called internally
**321** -> play method should be catched 

**To Reproduce (i.e. user steps, not code)**
1. set breakpoints in catch blocks line 75/324
2. run in debug mode

**Error messages**
-> please watch the log for "Unhandled Exception:"

**Expected behavior**
There should be at least one way to handle errors to reflect them to reactions.

**Smartphone (please complete the following information):**
 - iOS 16
 - Android 31

**Flutter SDK version**
```
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.3.5, on macOS 12.6 21G115 darwin-arm, locale de-DE)
[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] VS Code (version 1.72.1)
[✓] Connected device (5 available)
[✓] HTTP Host Availability

• No issues found!
```

**Additional context**
Some sweets for you! Thanks for your work.
![sweets](https://user-images.githubusercontent.com/3225274/200567583-de7b75cc-d5ed-4d6c-a9c2-70e33f555ca3.jpeg)

matecode:
@ryanheise 
I found out that:

- I used setAudioSource with preloading off
[...]

Issue #518, opened 2021-09-18
Just Audio does not cover all errors


sh4msi:
**Which API doesn't behave as documented, and how does it misbehave?**
This happened when I disconnected from the Internet and waited for the buffer to finish playing and try to play the next song.
In this case, I do not receive any errors by listening to playbackEventStream for error handling, while I receive UnknownHostException on the console.

**Minimal reproduction project**
If the explanation is not complete, I will do it

**To Reproduce (i.e. user steps, not code)**
1. Run code on the device
2. Playlist play
3. Internet disconnection
4. Wait like until the end of the playback of the buffer and auto go to the next song
5. See error in console, no listener reaction to error

**Error messages**
```
E/ExoPlayerImplInternal(20476): Playback error
E/ExoPlayerImplInternal(20476):   UnknownHostException (no network)
E/AudioPlayer(20476): TYPE_SOURCE: java.net.UnknownHostException: Unable to resolve host "cdn.***.com": No address associated with hostname
E/flutter (20476): [ERROR:flutter/lib/ui/ui_dart_state.cc(209)] Unhandled Exception: PlatformException(0, Source error, null, null)
```

**Expected behavior**
Listen to all errors

**Screenshots**
![error](https://user-images.githubusercontent.com/6229984/133886991-50d53ead-a9a9-4b6c-a8cc-9ea29a3fc422.jpg)

**Desktop (please complete the following information):**
 - OS: [Windows 10]
 - Browser [N/A]

**Smartphone (please complete the following information):**
 - Device: [Mi 9 SE]
 - OS: [Android 11]

**Flutter SDK version**
```
Doctor summary (to see all details, run flutter doctor -v):
[√] Flutter (Channel stable, 2.5.0, on Microsoft Windows [Version 10.0.17763.2183], locale en-US)
[√] Android toolchain - develop for Android devices (Android SDK version 29.0.3)
[√] Chrome - develop for the web
[√] Android Studio (version 4.2)
[√] VS Code (version 1.60.1)
[√] Connected device (3 available)

• No issues found!
```

ryanheise:
I need to know the minimal reproduction project to test these steps on. If the example exhibits the bug, please write "the example" under that section as per the instructions.

sh4msi:
I run and checked the official example (example_playlist.dart)
It is enough to proceed as I said

[...]

PR #1234, opened 2024-04-29
Workaround for issues #518 #870 

AErmek:
This is just a workaround for issues  #518 #870, not an official solution. And not tested on iOS devices.

I think these two issues are closely related to this [audio_service 987 PR](https://github.com/ryanheise/audio_service/pull/987) pull requests solution

### DIFF
--- a/just_audio_background/lib/just_audio_background.dart
+++ b/just_audio_background/lib/just_audio_background.dart
@@ -168,7 +168,12 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
   _JustAudioPlayer({required this.initRequest}) : super(initRequest.id) {
     _playerAudioHandler = _PlayerAudioHandler(
       initRequest,
-      addPlaybackEventMessageError: eventController.addError,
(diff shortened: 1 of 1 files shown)


just_audio_background/lib/just_audio_background.dart
```diff
@@ -166,7 +166,15 @@ class _JustAudioPlayer extends AudioPlayerPlatform {
   late final _PlayerAudioHandler _playerAudioHandler;
 
   _JustAudioPlayer({required this.initRequest}) : super(initRequest.id) {
-    _playerAudioHandler = _PlayerAudioHandler(initRequest);
+    _playerAudioHandler = _PlayerAudioHandler(
+      initRequest,
+      addPlaybackEventMessageError: (error, stackTrace) {
+        if (eventController.isClosed) {
```
inludese, line 172: This workaround will not work. 
eventController already closed, because disposePlayer called and after that player?.release();

```
Future<void> release() async {
    eventController.close();
    await _audioHandler.stop();
  }
```

Issue #440, opened 2021-07-05
Seek position of Flac audio actually leads to an earlier position


YukunXia:
**Which API doesn't behave as documented, and how does it misbehave?**
player.seek. 

The example uses seekBar widget, and dragging or clicking some position on the bar should move the progress of the audio playing to the corresponding position. However, for some files, eg. flac files longer than ~3min, the resultant position is earlier than the real position for 5~30s. So finally, the thumb of the seekBar has reached the end, but the audio is still on.

**Minimal reproduction project**
Provide a link here using one of two options:
https://github.com/YukunXia/just_audio/tree/bug_flac_duration

**To Reproduce (i.e. user steps, not code)**
1, Go to just_audio/example
2, `flutter run -d macos`
4, Click the `load audio` button
5, Click the play button
6. Click the seekBar at a position very close to the end, eg. 10 sec before the end of the duration.
7. Wait for the thumb of the seekBar to reach the end

**Error messages**

No error message

**Expected behavior**

The thumb of the seekBar has reached the end, but the audio is still on. For example, in the screenshot below, the audio is longer than the duration shown in the track for 30sec. However, if the song is played without being `seek`ed, this behavior won't happen. 

**Screenshots**

![image](https://user-images.githubusercontent.com/37142664/124432829-0e26c780-dd40-11eb-8400-78880f388ee1.png)

**Desktop (please complete the following information):**
 - macOS
 - Version: either the one in the repo or ^0.8.0

**Flutter SDK version**
```
flutter doctor                   
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel beta, 2.3.0-24.1.pre, on macOS 11.1 20C69 darwin-x64, locale en-US)
[!] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    ✗ cmdline-tools component is missing
      Run `path/to/sdkmanager --install "cmdline-tools;latest"`
      See https://developer.android.com/studio/command-line for more details.
    ✗ Android license status unknown.
      Run `flutter doctor --android-licenses` to accept the SDK licenses.
      See https://flutter.dev/docs/get-started/install/macos#android-setup for more
      details.
[✓] Xcode - develop for iOS and macOS
[✓] Chrome - develop for the web
[✓] Android Studio (version 4.2)
[✓] VS Code (version 1.57.1)
[✓] Connected device (2 available)
```

**Additional context**

[...]

PR #1238, opened 2024-05-04
[Fix] preferPreciseDurationAndTiming config unused in file Uri

canxin121:
Related: #440 

Already tested in my app, using flac with file Uri, the position after seek is correct.

### DIFF
--- a/just_audio/darwin/Classes/UriAudioSource.m
+++ b/just_audio/darwin/Classes/UriAudioSource.m
@@ -32,10 +32,20 @@
 
 - (IndexedPlayerItem *)createPlayerItem:(NSString *)uri {
     IndexedPlayerItem *item;
+    NSMutableDictionary *assetOptions = [[NSMutableDictionary alloc] init];
+    
+    if (_options != (id)[NSNull null]) {
+        NSDictionary *darwinOptions = _options[@"darwinAssetOptions"];
+        if (darwinOptions != (id)[NSNull null]) {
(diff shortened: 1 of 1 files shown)


canxin121:
@ryanheise Please review this pr.🤖

ryanheise:
I'm not sure if it is possible to test this. It looks like only the Objective C code is updated but there is no way to pass in this option from Dart.

canxin121:
>  I'm not sure if it is possible to test this. It looks like only the Objective C code is updated but there is no way to pass in this option from Dart.

> image https://github.com/canxin121/app_rhyme/actions/runs/8952578137
> 
> I used my fork to build my app in the lateset action, and tested it on my iphone13, works well.
> 
> The way to pass this configuration item remains exactly the same as before. Simply use:
> 
> ```dart
> ProgressiveAudioSource(
>     MusicUri,
>     tag: toMediaItem(),
>     options: const ProgressiveAudioSourceOptions(
>         darwinAssetOptions:
>             DarwinAssetOptions(preferPreciseDurationAndTiming: true)
>     )
> )
> ```
> 
> The function `createPlayerItem` previously ignored all configuration options for file type URI. This pull request simply reintroduces the use of the `preferPreciseDurationAndTiming` configuration option in file type URI.

ryanheise:
Ah, sorry I misunderstood at first glance. You aren't adding a new option, you're just moving the activation of the existing option to work for local files as well.

canxin121:
I recommend merging this PR, this simple change had no side effects during my testing

ryanheise:
Thank you for contributing, @canxin121 . This is now merged into `minor` (edit: removing response about comments/assertions which were for another PR).

PR #1136, opened 2023-12-08
Migrate to package:web

ryanheise:
Fixes #1127

To test, you need to comment/uncomment the relevant dependency overrides in `pubspec.yaml`.

PR #1241, opened 2024-05-16
Add cross origin setting to HTML5 AudioElement

danielwinkler:

Currently there is no support for headers in the just audio web player

One way to use authenticated audio sources on the web are using authentication cookies.

If the audio is served from a different origin than the website, this is considered a cross origin request.

In order to allow sending the credentials along with such a request, the audio element has a crossorigin attribute, see

[https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)

This PR attempts to allow setting this attribute on the `HTML5AudioPlayer` class.

Another valid option would be a constructer that accepts an `AudioElement`, such that the attribute can be set before passing it to just_audio.
The other option would be to expose the `_audioElement` with a getter.

**Additional context**

Given the current encapsulation, the `_audioElement` cannot be accessed, if we had a getter around it the following test could be added to the PR

```dart
import 'package:test/test.dart';
import 'package:just_audio_web/just_audio_web.dart';

void main() {
  group('Html5AudioPlayer', () {
    late Html5AudioPlayer player;

    setUp(() {
      player = Html5AudioPlayer(id: 'test_player');
    });

    test('should set anonymous', () {
      final request =
          SetCrossOriginRequest(crossOrigin: CrossOriginMessage.anonymous);
      player.setCrossOrigin(request);
      expect(player.audioElement.attributes['crossOrigin'], 'anonymous');
    });

    test('should set use credentials', () {
      final request =
          SetCrossOriginRequest(crossOrigin: CrossOriginMessage.useCredentials);
      player.setCrossOrigin(request);
      expect(player.audioElement.attributes['crossOrigin'], 'use-credentials');
    });
  });
}
```

### DIFF
--- a/just_audio/README.md
+++ b/just_audio/README.md
@@ -436,7 +436,7 @@ Please also consider pressing the thumbs up button at the top of [this page](htt
 | read from file                 | ✅      | ✅  | ✅    | ✅  | ✅      | ✅    |
 | read from asset                | ✅      | ✅  | ✅    | ✅  | ✅      | ✅    |
 | read from byte stream          | ✅      | ✅  | ✅    | ✅  | ✅      | ✅    |
-| request headers                | ✅      | ✅  | ✅    |     | ✅      | ✅    |
+| request headers                | ✅      | ✅  | ✅    | [^1] | ✅     | ✅    |
 | DASH                           | ✅      |     |       |     | ✅      | ✅    |
 | HLS                            | ✅      | ✅  | ✅    |     | ✅      | ✅    |
 | ICY metadata                   | ✅      | ✅  | ✅    |     |         |       |
@@ -455,6 +455,10 @@ Please also consider pressing the thumbs up button at the top of [this page](htt
 | skip silence                   | ✅      |     |       |     |         |       |
 | equalizer                      | ✅      |     |       |     |         | ✅    |
 | volume boost                   | ✅      |     |       |     |         | ✅    |
+
+[^1]: While headers cannot be set directly, cookies can be used to send information in the [Cookie header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie).
+See also `setWebCrossOrigin` to send cookies when loading audio files from a different origin (see [HTMLMediaElement crossOrigin](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/crossOrigin)).
+
 
 ## Experimental features
 
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -1196,6 +1196,18 @@ class AudioPlayer {
         usage: audioAttributes.usage.value));
   }
 
+  Future<void> setWebCrossOrigin(bool useCredentials) async {
+    if (_disposed) return;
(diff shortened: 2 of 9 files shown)


ryanheise:
Hi @danielwinkler 

Any new methods need to be added to the platform interface and then implemented in just_audio_web (even though this is a web specific API).

Another design could just be to just pass this through the existing headers API, and in the web implementation pull these headers out and translate them into a `setCrossOrigin` call.

I'm not suggesting that's better or worse, but those would both work.

danielwinkler:
Thanks for the feedback @ryanheise 

I added it to platform and prefixed `web`, I think the headers would be a bit confusing

To compile properly, I changed the ref in `pubspec.yaml` 

Do i have to split this to two PRs?

ryanheise:
One more thing is that this is just the backend side of the plugin, so a user of just_audio won't be able to use this feature until there is a method in just_audio's AudioPlayer class. It can be implemented by simply delegating to the platform interface.

Regarding the temporary change to `pubspec.yaml`, keep that change locally, but commit the different `pubspec.yaml` files as they should appear in the final release. I'm not a fan of the complications the federated plugin model causes for the release process, and a two PR approach has been considered before, but I'm not currently doing that.

danielwinkler:
Thanks a lot for the guidance @ryanheise 

I increased the version numbers everywhere starting from the platform interface

on the AudioPlayer level i exposed the web cors functionality as `setWebCrossOrigin(bool useCredentials)`

ryanheise:
I think I will also be merging #1136 first given that wasm support has now entered flutter stable, and this may result in some conflicts unfortunately. I will let you know once I have done that.

danielwinkler:
Sounds good!

Cheers,
daniel

ryanheise ***@***.***> schrieb am Do. 16. Mai 2024 um 16:55:

ryanheise:
I've merged the other PR into minor now. By chance it looks like there are no merge conflicts, but you might want to test it to be sure. Note that the other PR already bumped just_audio_web to 0.4.10 and that's now released, so yours would need to be bumped to 0.4.11.

danielwinkler:
Thanks for the update @ryanheise 

I updated the versions and added information to the README and changelog

ryanheise:
I'll working through this now. I found some issues, but I'll sort them out. Mainly naming and docs.

ryanheise:
I've made my changes in branch `feature/bump_web`. There are some name changes, and I've made the parameter in the main plugin into an enum to more directly reflect the underlying web API as well as making the parameter nullable so that you can "unset" the crossorigin attribute.

Not sure if you will see this in time but I will be publishing this imminently. The main thing that can't be changed is the API (once it's published), but if there are any bugs in the implementation of the platform interface, that can be fixed if you happen to find that what I've changed doesn't work for you.

ryanheise:
This has now been released on pub.dev. Although it remains on the bump_web branch for now, until I fix an issue in just_audio_background.

ryanheise:
I've updated just_audio_background to support this method. It is just released, and this branch is now merged. Thanks for your contribution. If you get a chance, you might want to review the changes I've made to understand what has changed.

PR #1242, opened 2024-05-16
WIP: trying to replace dart:html with package:web for wasm

jarrodcolburn:
WIP trying to rewrite to remove dart:html so it can be compiled with wasm


ryanheise:
Duplicate of #1136

PR #1247, opened 2024-05-27
added refresh token mechanism

MrGhostXUz:
In our music player app, a client need to have Authorization headers. When I gave ConcatenatingAudioSource, the player was stuck after some time, when the given token is expired. So I decided to add inbuilt mechanism to refresh credentials by adding some optional fields. I am sure that I didn't break any existing features.

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -2077,6 +2077,8 @@ class _ProxyHttpServer {
   Uri addUriAudioSource(UriAudioSource source) {
     final uri = source.uri;
     final headers = <String, String>{};
+    final refreshCredentials = source.refreshCredentials;
+    final getAuthHeaders = source.getAuthHeaders;
     if (source.headers != null) {
       headers.addAll(source.headers!.cast<String, String>());
     }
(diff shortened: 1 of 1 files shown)


ryanheise:
This solution seems specific to the proxy (i.e. won't work with `useProxyForRequestHeaders = false`). In that case, have you considered doing this outside the plugin with a custom stream audio source?

MrGhostXUz:
> This solution seems specific to the proxy (i.e. won't work with `useProxyForRequestHeaders = false`). In that case, have you considered doing this outside the plugin with a custom stream audio source?

Yes, the solution is specific for ConcatenatingAudioSource in other word for something like playlists. I think, if someone needs this auth mechanism with `useProxyForRequestHeaders = false`, providing concrete cases, then I will expand my solution. Until then, I think it is time-wasting.

ryanheise:
I want to avoid highly specific solutions, so I am not going to be in a rush to make design decisions such as this one. For the library itself, there is a preference to more general APIs that provide the flexibility to do various things. Is StreamAudioSource enough to allow an implementation within the app? If not, let me know so that we can design another abstraction that would be powerful enough to allow the app to build such things on top of it.

PR #1268, opened 2024-06-22
Fix  issue #594 by force starting proxy on load

AmmarZY:
I have updated _load method to force staring proxy before loading the request, this is a workaround solution that fix the issue.


ryanheise:
That is an interesting approach, although we also don't want to end up with boundless new proxies being started up without any of them ever being stopped, so this idea may need some further development.

PR #1283, opened 2024-07-08
Fix RXDart version not compatible with the lasted 0.28.0

prologikus:
Fix RXDart version not compatible with the lasted 0.28.0


ryanheise:
It looks like I accidentally fixed this myself instead of merging your PR (had I been paying attention I would have used your PR!)

Issue #881, opened 2022-11-23
M4A not buffering on iOS, macOS


aguilaair:
**Which API doesn't behave as documented, and how does it misbehave?**
Name here the specific methods or fields that are not behaving as documented, and explain clearly what is happening.

**Minimal reproduction project**
Simply replace the the initMethod of the `just_audio` example with an .m4a audio file like:

https://iwwhjxplzmgcmpmhpqt.supabase.co/storage/v1/object/public/email/instrumental.m4a?t=2022-11-23T01%3A09%3A41.568Z

Fork: https://github.com/aguilaair/just_audio/tree/minor/just_audio/example

**To Reproduce (i.e. user steps, not code)**
Open example and wait for buffering to happen, will fully buffer

**Expected behavior**
the .m4a file should buffer instead of downloading the whole thing before playing

**Desktop (please complete the following information):**
 - OS: macOS Ventura
 - Browser: Safari, native app

**Smartphone (please complete the following information):**
 - Device: iPhone 14 Pro Max
 - OS: iOS 16 Dev Beta 3

**Flutter SDK version**
```
[✓] Flutter (Channel stable, 3.3.8, on macOS 13.0 22A380 darwin-arm, locale en-US)
    • Flutter version 3.3.8 on channel stable at /Users/emorenoa/fvm/versions/stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 52b3dc25f6 (2 weeks ago), 2022-11-09 12:09:26 +0800
    • Engine revision 857bd6b74c
    • Dart version 2.18.4
    • DevTools version 2.15.0

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/emorenoa/Library/Android/sdk
    • Platform android-33, build-tools 33.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.12+0-b1504.28-7817840)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.1)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14B47b
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2021.2)
[...]

PR #1296, opened 2024-07-26
[just_audio] Ignore acceptEncodingHeader from proxy server

dfdgsdfg:
# Ignore accept-encoding Header

If the origin response has the `accept-encoding` HTTP header with `useProxyForRequestHeaders: true`, just ignore it.

Because the default HttpServer does not compress responses with GZIP, but the header has `accept-encoding`, the native encoder throws an error.

Another approach is to add the `autoCompress: true` option, but I'm not sure it covers all cases.

## References
- maybe relatated #881 ?
- https://api.dart.dev/stable/3.4.4/dart-io/HttpServer/autoCompress.html

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3330,10 +3330,15 @@ _ProxyHandler _proxyHandlerForUri(
     String? host;
     try {
       final requestHeaders = <String, String>{};
-      request.headers
-          .forEach((name, value) => requestHeaders[name] = value.join(', '));
+      request.headers.forEach((name, value) {
+        if (name == HttpHeaders.acceptEncodingHeader) return;
+        requestHeaders[name] = value.join(', ');
+      });
       // write supplied headers last (to ensure supplied headers aren't overwritten)
-      headers?.forEach((name, value) => requestHeaders[name] = value);
+      headers?.forEach((name, value) {
+        if (name == HttpHeaders.acceptEncodingHeader) return;
+        requestHeaders[name] = value;
+      });
       final originRequest =
           await _getUrl(client, redirectedUri ?? uri, headers: requestHeaders);
       host = originRequest.headers.value(HttpHeaders.hostHeader);


ryanheise:
Apologies for the delay, I had computer issues. This looks interesting, I shall review on the weekend. Although if the proxy is just copying the exact headers, wouldn't it be that #881 would be an issue even without the proxy?

dfdgsdfg:
It seems the accept-encoding header is a fairly robust feature, according to the standard.
[http 1.1 accept-encoding header rfc](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.4)

Issue 881 also includes an accept-encoding and gzip-related option. So, in my opinion, it’s fine to proceed without the proxy, perhaps?

Unfortunately, the URL provided in issue 881 is no longer working, so I couldn’t run the test. However, similar to the example, our m3u8 static media server also goes through AWS CloudFront, with Gzip enabled in the responses.

[Cloudfront gzip](https://aws.amazon.com/ko/blogs/aws/new-gzip-compression-support-for-amazon-cloudfront/)

For Android, I disable the proxy with `useProxyForRequestHeaders: false`, and it works well. However, I haven’t tested it on iOS.

ryanheise:
Note that your change only affects the proxy scenario, and we have no control over what the native os decides to do when we are not using the proxy (hence the question about whether you have observed the issue even without the proxy)

ryanheise:
Do you have a URL I can test this with?

Also, my preference would be to aim to do a direct passthrough, so with regards to compression, I would prefer that if the native player says it accepts a certain encoding, and the server actually sends a certain encoding, I would like to preserve the encoding that the server sends and just forward that back to the player again so that the middleman is unnoticeable. Maybe that means setting `autoUncompress = false` on the client and leaving the headers alone. Thoughts?

ryanheise:
I'd like to test this hypothesis if you have a URL to test on. Am I understanding it correctly that this issue occurs when the server serves a gzipped response because the client declared that it would accept it?

I have never encountered a server that actually gzips audio responses, so it's a bit difficult to reproduce, if that's the case. But if it is the case, I would like to test the approach described in my previous comment.

Issue #980, opened 2023-05-03
ExoPlayer is deprecated, Migrate to Media3!


akshdeep-singh:
**Is your feature request related to a problem? Please describe.**
The standalone `ExoPlayer` project, with package name `com.google.android.exoplayer2`, will soon be discontinued, and future updates will be published in `Media3`

**Describe the solution you'd like**
Migrate to `Media3` - [Migration Guide](https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide)

**Describe alternatives you've considered**
NA

**Additional context**
Read Blog for details - [Media3 is ready to play!](https://android-developers.googleblog.com/2023/03/media3-is-ready-to-play.html)

nateshmbhat:
@ryanheise  any update on this ? all major players are moving to media3 exoplayer. would be really awesome if just_audio also migrates ! We're using this just_audio in many of our applications.

PR #1325, opened 2024-09-12
Fix issue #980 migrate from ExoPlayer to Media3

hansvdwd:
Updated just_audio from ExoPlayer (deprecated) to Media3.


ryanheise:
Thanks! This is definitely appreciated. Before merging, I'd like to coordinate this with audio_session and maybe audio_service, although migrating audio_service to media3 is a much bigger ordeal that I'm not looking forward to, and perhaps it is not necessary if audio_service "as is" still works with your PR. However, migrating audio_session should be a much simpler task.

Can you format your ChangeLog entry as per [CONTRIBUTING.md](https://github.com/ryanheise/just_audio/blob/b5fad794dee004a9cb92804fce6fcb12d025a0f8/CONTRIBUTING.md)?

just_audio/android/build.gradle
```diff
@@ -48,11 +48,10 @@ android {
     }
 
     dependencies {
-        def exoplayer_version = "2.18.7"
-        implementation "com.google.android.exoplayer:exoplayer-core:$exoplayer_version"
-        implementation "com.google.android.exoplayer:exoplayer-dash:$exoplayer_version"
-        implementation "com.google.android.exoplayer:exoplayer-hls:$exoplayer_version"
-        implementation "com.google.android.exoplayer:exoplayer-smoothstreaming:$exoplayer_version"
+        def exoplayer_version = "1.0.2"
```
nateshmbhat, line 51: please consider using the latest [1.4.1 version of media3 exo player](https://github.com/androidx/media/releases)

nateshmbhat (changes requested):

just_audio/android/build.gradle
hansvdwd: Thank you for the feedback. And good point, I had a quick look at using the latest version and that is not so trivial unfortunately. It would be easy to switch to version 1.1.1 but anything newer would require several code changes. The class _ConcatenatingMediaSource_ was deprecated and removed from the media3 library. I will have a look at what changes are required to jump to the latest version. But it might also be worth considering to split it in two separate pull requests.

ryanheise: I'd like to say we can just do an intermediate release with version 1.1.1 and then work on 1.4.1, but  video_player uses 1.4.1, and we have seen version conflicts occur before in apps that use both plugins when the exoplayer dependencies aren't in sync.

Let me take a look at this and see if I can tackle 1.4.1 using this PR as a starting point...

suyie001:
Any progress?

ryanheise:
Sorry, my previous computer broke down so I've been without access for a while, but I'm back up and running now. I have a bit to catch up on, but 1.4.1 is on my list.

ryanheise:
Today I worked on the 1.4.1 migration. i fully expected to have to rewrite the media source code, but as it turned out, those classes are still there, so for the moment I've left the code as is with deprecation warning suppressed.

The main effort then was to rewrite the offload scheduling code.

See PR #1377 . Would anyone like to test this?

Note that the way it "should" work is that if offload scheduling is enabled, it may prevent you from changing the speed and may disrupt gapless playback depending on device support. If offload scheduling is disabled (the default), then actually ExoPlayer will try to still enable offload scheduling if possible, but only if it won't disrupt gapless or speed changes. That's a slight deviation from the previous behaviour, but tries to give the best of both worlds without breaking anything. (Hopefully I didn't break anything!)

ryanheise:
I've now merged #1377 but to ensure that both PRs are credited, I merged my PR over the top of this one so both are in the commit log and CHANGELOG.

PR #1327, opened 2024-09-15
Update README.md

arolleaguekeng:
The named parameter 'second' isn't defined. 
correct parameter is seconds


ryanheise:
Thanks for noticing that. I've corrected a minor alignment issue and merged.

PR #1335, opened 2024-09-23
fix when play music in background thread, play state will change wrong

zhushenwudi:
https://github.com/user-attachments/assets/dbea5640-4d21-4a84-ae19-c5b63056893c

screenshot-20240924-070517

When I play with other music player, click a home widget to play with just_audio, it will throw a exception from native core.
At this point, we should catch the exception and change the playback status to stop.
Or just_audio can provide the ability to interrupt other audio, I think both are possible.

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -938,26 +938,34 @@ class AudioPlayer {
     _playbackEventSubject.add(_playbackEvent);
     final playCompleter = Completer<dynamic>();
     final audioSession = await AudioSession.instance;
-    if (!_handleAudioSessionActivation || await audioSession.setActive(true)) {
-      if (!playing) return;
-      // TODO: rewrite this to more cleanly handle simultaneous load/play
-      // requests which each may result in platform play requests.
-      final requireActive = _audioSource != null;
-      if (requireActive) {
-        if (_active) {
-          // If the native platform is already active, send it a play request.
-          // NOTE: If a load() request happens simultaneously, this may result
-          // in two play requests being sent. The platform implementation should
-          // ignore the second play request since it is already playing.
-          _sendPlayRequest(await _platform, playCompleter);
-        } else {
-          // If the native platform wasn't already active, activating it will
-          // implicitly restore the playing state and send a play request.
-          _setPlatformActive(true, playCompleter: playCompleter)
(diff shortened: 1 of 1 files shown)


ryanheise:
Are you able to link this to a bug report? This will help me to reproduce the issue and test your fix. Can you reproduce this without a home widget and maybe just the media notification carousel?

zhushenwudi:
> Are you able to link this to a bug report? This will help me to reproduce the issue and test your fix. Can you reproduce this without a home widget and maybe just the media notification carousel?

There is no specific project link, and restoring this scenario is quite complicated. Currently, I have been using the fork repository as a dependency because I have been unable to use the latest version of just audio to meet my needs for a long time. Because the new version of Exoplayer is not compatible with FLAC third-party library jar packages (functions have been abandoned), I have to stay in the old version. At present, submitting a pull request is only for reference, and you may need to evaluate whether this part is necessary.

ryanheise:
I'd like to investigate, but I would need your help to reproduce it in order to be sure the correct solution is implemented.

ExoPlayer is a separate matter (Android side) and actually my goal is to move forward upgrading to an even newer version since the currently used version of ExoPlayer has been deprecated.

zhushenwudi:
Ok, if I have time I will find the easiest demo program to reproduce this scenario. I am also looking forward to the upgrade of just_audio to media3, but I still don't know if it can meet my current needs. If there is an update in the future, I will also try the first time.

ryanheise:
The ExoPlayer code has been partially upgraded on `minor`. Any instructions on how to reproduce the bug?

zhushenwudi:
I am sorry that I have been working on other projects recently and time is tight. However, I found time to integrate the media3 branch of just_audio before. I can only say that the effect seems to be the same as that of exoplayer2, with no improvement. When the song itself is defective, media3 still can't skip the wrong data and continue playing backwards. There will be a switch to the next song, so I'll still use my own version of the fork repository. This has had a significant impact on the follow-up feedback.

ryanheise:
Yes, I wouldn't expect media3 to change anything in this respect. There is still another bit of work to do with media3 later, but I don't expect that to affect your issue either.

So the main blocker for me is that I can't investigate your issue since I don't have a means to reproduce it and verify that the correct solution has been implemented.

PR #1361, opened 2024-11-12
Update just_audio.dart

ZenW1:
udpate random secure


ryanheise:
Hi ZenW1, ironically this change is not secure/stable. Two problems:

1. You should not create a Random instance on each invocation. You're ignoring the global definition where it is instantiated.
2. The documentation states this can throw an `UnsupportedError` if the platform does not support this.

ryanheise:
3rd problem: the `secure` is intended for cryptographic purposes. This plugin uses `Random` to shuffle a playlist, which is not a cryptographic purpose. If you have a strange need for this, the shuffle order API is extensible, you can implement this within your own app.

PR #1369, opened 2024-11-26
feat: Add setDeviceId method on web

dganzella:
Added setDeviceId method, so the output speaker (HTML5AudioElement sink ID) can be set -- so developer can select which device to play sound instead of default

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -1529,8 +1529,8 @@ class AudioPlayer {
   }
 
   /// Sets a specific device output id, null for default
-  void setDeviceId(String? id) {
-    _platform.setDeviceId(id);
+  void setDeviceId(String? id) async {
+    (await _platform).setDeviceId(id);
   }
 
   /// Dispose of the given platform.
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -1529,8 +1529,8 @@ class AudioPlayer {
   }
 
   /// Sets a specific device output id, null for default
-  void setDeviceId(String? id) async {
-    (await _platform).setDeviceId(id);
+  Future<void> setDeviceId(String? id) async {
+    await (await _platform).setDeviceId(id);
   }
 
   /// Dispose of the given platform.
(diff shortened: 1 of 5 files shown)


ryanheise:
I would suggest renaming this to `setWebDeviceId` following the convention of `setWebCrossOrigin` since it is a web-specific method. Or given that it's specific to the web API, maybe `setWebSinkId` so there is no confusion about what it represents? If you're thinking of generic naming so that multiple platforms can implement it in a different way, I think there are too many platform differences in the way audio routing APIs are designed that this would not work. Maybe there is a similar notion in Android, but in just_audio it is again named `androidAudioSessionId` as it is unlikely Android's API works in a similar way to other platforms (for example, you don't set it, you listen to it and then pass it into separate APIs for routing.)

dganzella:
> I would suggest renaming this to `setWebDeviceId` following the convention of `setWebCrossOrigin` since it is a web-specific method. Or given that it's specific to the web API, maybe `setWebSinkId` so there is no confusion about what it represents? If you're thinking of generic naming so that multiple platforms can implement it in a different way, I think there are too many platform differences in the way audio routing APIs are designed that this would not work. Maybe there is a similar notion in Android, but in just_audio it is again named `androidAudioSessionId` as it is unlikely Android's API works in a similar way to other platforms (for example, you don't set it, you listen to it and then pass it into separate APIs for routing.)

k, I will fix it this week and send to you

ryanheise:
Did you have an opinion on `setWebDeviceId` vs `setWebSinkId`? I think "sink" is more faithful to the original API.

dganzella:
> Did you have an opinion on `setWebDeviceId` vs `setWebSinkId`? I think "sink" is more faithful to the original API.

Yes, I also think setWebSinkId is better!

ryanheise:
Then let's go with the `setWebSinkId` naming, and then I can merge and release.

dganzella:
ok @ryanheise I moved this PR to another, because I needed to revert the pubspec.yaml but did not want to change  the code that is working in production to stop compiling

https://github.com/ryanheise/just_audio/pull/1378

PR #1377, opened 2024-12-15
Migrate to media3 ExoPlayer 1.4.1 on Android.

ryanheise:
Migrate to the latest ExoPlayer while maintaining use of the deprecated media source API. Rewriting that will be future work.

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 0.9.43
 
 * Fix NPE in load on iOS/macOS.
+* Migrate to media3 ExoPlayer 1.4.1 on Android.
 
 ## 0.9.42
 
--- a/just_audio/android/build.gradle
+++ b/just_audio/android/build.gradle
@@ -48,11 +48,11 @@ android {
     }
 
     dependencies {
-        def exoplayer_version = "2.18.7"
-        implementation "com.google.android.exoplayer:exoplayer-core:$exoplayer_version"
-        implementation "com.google.android.exoplayer:exoplayer-dash:$exoplayer_version"
-        implementation "com.google.android.exoplayer:exoplayer-hls:$exoplayer_version"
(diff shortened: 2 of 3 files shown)


Chaphasilor:
@ryanheise I plugged this PR into [Finamp](https://github.com/jmshrv/finamp) and gave it a try, everything seemed to work normally. Tried various codecs and playback sources, and also changing playback speed.  
I tested on an ASUS Zenfone 10 running Android 14.  
We're using a `ConcatenatingMediaSource` for our queuing system, and all of the related functionality seemed to work.

Anything in particular to test or look out for?  
Really excited about this, hopefully the new ExoPlayer will be even more reliable than the old version!

ryanheise:
Thanks for testing, @Chaphasilor . The main things that could be affected by update are speed changing, gapless playback and offload scheduling. I've tried to hide the underlying changes so that the plugin behaves as before, so ideally there are no behavioural changes.

This is a relatively simple change, but at some point I will need to do a more significant change to remove the deprecated ExoPlayer API calls. Due to the deprecation of certain APIs, I'd be inclined to simplify the just_audio API accordingly. Rather than having a separate `ClippingAudioSource` class, I may just have a `clipStart` and `clipEnd` field on all audio sources. And then instead of having a separate `ConcatenatingAudioSource`, I may just make the player instance itself encapsulate the sequence of audio sources.

Chaphasilor:
Sorry for the late reply. I'll have to play around with forced offloading to confirm that everything is working as expected. Aside from that, still no issues.  

A simplified API with less specialized audio sources sounds nice. I'd be especially keen on getting more manual control over the queue, right now we're doing some serious gymnastics for synchronizing the queue between our app and the media source, including a custom shuffle order. The tag functionality of `MediaItems` really helps here. But that's a different topic, one step at a time :D

ryanheise:
This has been merged.

Chaphasilor:
Sorry for never responding. We'll test this in the next beta update and will try to report any issues :)

PR #1378, opened 2024-12-16
feat: add setWebSinkId 

dganzella:
Added setWebSinkId method, so the output speaker (HTML5AudioElement sink ID) can be set -- so developer can select which device to play sound instead of default


ryanheise:
From the documentation, I can't confirm that `setSinkId` accepts an empty string argument, or at least I've found hints that it might be, but no direct confirmation in the documentation of `setSinkId` itself. So I'm not sure if it's well-defined to make `setWebSinkid` accept `null` which then gets translated to an empty string. Do you have any information on this?

dganzella:
> From the documentation, I can't confirm that `setSinkId` accepts an empty string argument, or at least I've found hints that it might be, but no direct confirmation in the documentation of `setSinkId` itself. So I'm not sure if it's well-defined to make `setWebSinkid` accept `null` which then gets translated to an empty string. Do you have any information on this?

Yes

https://developer.chrome.com/blog/audiocontext-setsinkid/

thar said, web is always misterious, we really depend on the developers of the browsers to implement things the same way, which not always happens

I tried  on Chrome and it works, as suggested by their blog post

ryanheise:
Might not that be Chrome specific and not standardised yet? I would prefer to have official documentation on the actual standard pointing to that, or otherwise perhaps just define the Dart API to accept a non-nullable String parameter and leave it up to the user to try their luck at browser-specific behaviour of passing in an empty string. But I would still like to know if it makes sense to hard code null to mean empty string if that is or is not the w3c standard.

dganzella:
> Might not that be Chrome specific and not standardised yet? I would prefer to have official documentation on the actual standard pointing to that, or otherwise perhaps just define the Dart API to accept a non-nullable String parameter and leave it up to the user to try their luck at browser-specific behaviour of passing in an empty string. But I would still like to know if it makes sense to hard code null to mean empty string if that is or is not the w3c standard.

ok, I can change to non nullable string, no problem

but waiting for official documentation you will die waiting haha

ryanheise:
Then I think just mirroring the underlying web API exactly will be the most straightforward way forward, so just passing through whatever is the behaviour of the underlying web browser. Since the underlying API does not accept null parameters, just_audio should not introduce this. If you agree, I can take care of it on my end before merging.

dganzella:
sure, I agree :)

ryanheise:
Can you please review the changes in #1399 ? Specifically, my changes are in this commit: 4ac4ab1b45bdece2146adc1f08432e775d40ec78

Issue #1119, opened 2023-11-10
should customize connect and read timeout when the audio file is large or network is lag sometime


clevertension:
**Which API doesn't behave as documented, and how does it misbehave?**
when I play more than 70Mb mp3 file in remote url, the AudioPlayer always retry after 5 seconds

**Minimal reproduction project**

**To Reproduce (i.e. user steps, not code)**
You can find a large mp3 file in remote url, and play with it, if the network is lag more than 5 seconds, it will fail to play

**Error messages**
it seems no error message

```
If applicable, copy & paste error message here, within the triple quotes to preserve formatting.
```

**Expected behavior**

**Screenshots**
![image](https://github.com/ryanheise/just_audio/assets/346723/6cefa249-3a4b-4d29-9e92-cd9d16bff100)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**
 - Device: [e.g. XiaoMi]
 - OS: [e.g. Android 12]

**Flutter SDK version**
```
insert output of "flutter doctor" here
```
[√] Flutter (Channel stable, 3.13.8, on Microsoft Windows [版本 10.0.17763.4974], locale zh-CN)
[√] Windows Version (Installed version of Windows is version 10 or higher)
[!] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    ! Some Android licenses not accepted. To resolve this, run: flutter doctor --android-licenses
[√] Chrome - develop for the web
[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.7.4)
[!] Android Studio (version 2022.2)
    X Unable to find bundled Java version.
[√] Android Studio (version 2022.3)
[√] IntelliJ IDEA Ultimate Edition (version 2022.3)
[√] Connected device (4 available)
[√] Network resources

**Additional context**
please see the code snapshot above, if i add the read timeout option, the problem is solved, so we can provide such option

clevertension:
I modify file **com.ryanheise.just_audio.AudioPlayer.java**, line 694

ryanheise:
Can you please provide a reproducible example that I can run?

kaciula:
@ryanheise I am also interested in having a way to set the timeouts. I get hundreds of reports of 

`"(-1001) The request timed out"`

To reproduce this, you can throttle the network connection to be slower. On iPhone, go to Settings -> Developer -> Network Link Conditioner and choose a slow Internet profile (I have one called Very Bad Network)

[...]

PR #1382, opened 2024-12-26
Add timeout option to uri requests

ctedgar:
Hi @ryanheise,

This is a pull request to address #1119, it includes the following changes:

1. just_audio_platform_interface: Updated the implementations of UriAudioSourceMessage to include the optional timeoutMillis parameter.
2. just_audio: 
-  Added timeout parameters to UriAudioSource and StreamAudioSource implementations.
-  Updated AudioPlayer.java to expect the optional timeout parameter and respect it if set. 
-  Added a test case under _'load error'_ to simulate the timeout error and to also validate the plumbing of the new parameter. (i.e. assert that the timeout property is making it down to the platform message)

I realize there may be a need to implement the same functionality for all other platforms. I've just carried out the changes for Android as it's what I could test comfortably.

I added change notes and bumped the versions both just_audio and just_audio_platform_interface though I suppose just_audio_platform_interface would need to be published first followed by just_audio, if these changes are accepted ofcourse.  

Note for local testing: Update just_audio_platform_interface/just_audio_web dependency in just_audio pubspec to point to local folder

Regards,
Cris

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.9.47
+
+* Add timeoutMillis parameter to Uri/LockCaching/Stream AudioSources (@ctedgar).
+    * Android platform implementation completed
+
 ## 0.9.46
 
 * Fix SwiftPM support on macOS.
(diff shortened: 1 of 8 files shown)


ctedgar:
Had not accounted LockCachingAudioSource and the URI proxied requests, so added similar timeout overrides LockCaching requests and for proxied requests.

Regards,
Cris

ctedgar:
Hi @ryanheise,

Just a reminder in case this was forgotten. Did you get a chance to look into this again? Any suggestions? If not I can fix conflicts on change log files to bring it up-to-date.

Thank you, 
Cris

ctedgar:
Hi @ryanheise, 

Another reminder in this :) 

Thanks,
Cris

PR #1416, opened 2025-03-03
VulnerabilityFix - Use secure Random

smiron-bnext:
The binary can use the following insecure random functions: Random

Replace Random with functions that use cryptographically secure random number generation (CSPRNG).
For iOS and Flutter, use Random.secure() (in Dart), which utilizes secure system sources such as /dev/urandom.

https://learn.snyk.io/lesson/insecure-randomness/
https://github.com/MobSF/owasp-mstg/blob/master/Document/0x04g-Testing-Cryptography.md#weak-random-number-
generators
https://cwe.mitre.org/data/definitions/330.html

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3506,7 +3506,7 @@ class DefaultShuffleOrder extends ShuffleOrder {
   @override
   final indices = <int>[];
 
-  DefaultShuffleOrder({Random? random}) : _random = random ?? Random();
+  DefaultShuffleOrder({Random? random}) : _random = random ?? Random.secure();
 
   @override
   void shuffle({int? initialIndex}) {


ryanheise:
`Random.secure()` is intended for cryptographic purposes. just_audio uses `Random` to shuffle a playlist which is not a cryptographic purpose. The default shuffle order implementation already allows you to pass in your own custom `Random` instance, so this PR is unnecessary. And you can also create your own subclass to further control random generation.

This is also not the first time someone has submitted such a PR (#1361 ) so I'm wondering why you feel this is necessary given that this plugin has nothing to do with cryptography?

ryanheise:
Reading my comments on the previous PR, the documentation says that `Random.secure()` can also throw an `UnsupportedError` so this would not be a satisfactory implementation even if I were to be convinced that we need a cryptographically strong random number generator in order to just generate shuffled playlists.

ryanheise:
Can you please explain what exactly is the security vulnerability here? If not, the PR should be closed. But since you are the second person to try to fix this "vulnerability", I am scratching my head trying to understand the reasoning. :shrug:

smiron-bnext:
Insecure randomness refers to random numbers or data that do not have enough entropy, or randomness, to be considered truly random. This occurs when a function that is expected to return a random value actually has a predictable output, allowing an attacker to compromise security.

More than anything, it would just be about using a secure way to generate the random number, regardless of its use, since during penetration testing, this vulnerability is marked as a priority. It would be very kind of you to comply with these security measures, even in something very basic.

In my first comment, you will find the references.

ryanheise:
> Insecure randomness refers to random numbers or data that do not have enough entropy, or randomness, to be considered truly random. This occurs when a function that is expected to return a random value actually has a predictable output, allowing an attacker to compromise security.

You're insulting my intelligence. I did not ask you to explain what insecure randomness refers to, OBVIOUSLY I already know what it is. This should be clear. (I am a university computer science lecturer -- I understand this topic).

My question to you was: where's the vulnerability specifically in the context of THIS plugin? Not in generic terms, but how exactly is it that you're seriously concerned about the shuffle order of a playlist? Are you WORRIED that someone is going discover or predict the order of my playlist or intercept it or modify it? Of course not. Those things are inconsequential to security in the context of THIS plugin.

And as I pointed out, your PR will ironically make this plugin LESS reliable because it can potentially throw an exception and you did not handle the exception. You are blindly replacing one API with another API without even understanding the implications of the change, and without even being able to explain precisely HOW this will make the plugin more secure, given that this plugin has nothing to do with security in the first place.

And, you have ignored the fact that it is already possible for the user to pass in their own random generator to the constructor parameter if that's what they want.

You are not answering any of my questions, you are just blindly giving a boiler plate answer as if you understand nothing about the specifics of this plugin.

The only thing I can think of to explain this strange behaviour is there may be something like Hacktoberfest going on where developers are being given financial or other incentives by some sponsor to submit PRs on GitHub, and unfortunately this attracts a lot of useless and poor quality PRs like this one which would actually break the plugin (in your case, by not catching and handling the exception, ironically in the purported name of security). All I can say is, no thanks.

If you can't explain how playlist shuffling presents an actual risk as it currently is implemented, and why this needs to be addressed, other than your abstract wishy washy generic comments about "regardless of its use", then no thanks, please don't waste people's time.

I will close this issue, I don't expect you'll actually have an answer to any of the above anyway. But if you think you have an answer, go ahead. I don't need you to teach me about security, I just need direct answers to the questions that you did not answer.

PR #1436, opened 2025-04-14
Allow audio_session 0.2.0

hpoul:
audio_session 0.2.x does not introduce any breaking changes to the API, but only to the iOS setup. Hence no code changes are required.

### DIFF
--- a/just_audio/pubspec.yaml
+++ b/just_audio/pubspec.yaml
@@ -20,7 +20,7 @@ dependencies:
   just_audio_web: ^0.4.14
   # just_audio_web:
   #   path: ../just_audio_web
-  audio_session: ^0.1.24
+  audio_session: ">=0.1.24 <0.3.0"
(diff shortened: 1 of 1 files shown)


ryanheise:
This is still technically a breaking change. iOS apps that depend on AUDIO_SESSION_MICROPHONE being set to 1 yet assume the default documented value will break with this change, and it would require a code change to explicitly set it.

hpoul:
It's a breaking change for `audio_session`, not for `just_audio` .. 🤷 and just_audio does not record audio (afaik?), so disabling it wouldn't hurt anyone not using it directly.. and everyone using it directly would hopefully include it in their pubspec anyway.. so would have to update it manually?

ryanheise:
That reasoning is fair, although can you rebase this to `feature/playlist`? This is the next just_audio release, which will itself be a breaking change (whether or not the audio_session constraint change is breaking).

PR #1452, opened 2025-05-02
Fix stop() not broadcasting idle state

rserro:
When the audio player is stopped, the device player is not dismissed because idle state is not being broadcasted.

### DIFF
--- a/just_audio_background/lib/just_audio_background.dart
+++ b/just_audio_background/lib/just_audio_background.dart
@@ -73,19 +73,11 @@ class JustAudioBackground {
     );
   }
 
-  /// Clear the current device playback.
-  /// On iOS, this has the effect to dismiss the device player.
-  static Future<void> clearPlayback() async {
-    await _audioHandler.stop();
+  /// Ensure the player state is broadcasted to the device.
+  static Future<void> broadcastState() async {
     if (_audioHandler.inner is _PlayerAudioHandler) {
       final playerHandler = _audioHandler.inner as _PlayerAudioHandler;
-      playerHandler._justAudioEvent = playerHandler._justAudioEvent.copyWith(
-        processingState: ProcessingStateMessage.idle,
-      );
-      playerHandler._playing = false;
       playerHandler._broadcastState();
-      playerHandler.updateQueue([]);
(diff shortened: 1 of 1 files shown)


rserro:
@ryanheise Please have a look at this. It is an important missing feature.

ryanheise:
It seems like all of the APIs you're using in this PR are already accessible to the app - e.g. `player.clearAudioSources()` and `player.stop()`. If those are not working as intended, maybe that is a better starting point to investigate rather than changing the API surface.

rserro:
True but player.stop() and player.clearAudioSources() do not achieve the effect of dismissing the device player. I have updated the PR to narrow down the code only to the broadcast action. If player.stop() is called before and then JustAudioBackground.broadcastState(), it now works. I could not understand why player.stop() and player.clearAudioSources() do not achieve this effect, as they seem to broadcast already. Also, most of the code for JustAudioBackground implementation are kept _private, which does not help in reusing already implemented codes.

ryanheise:
I think you misunderstood my previous comment: if the existing API doesn't work, you can fix it without changing the surface API. Your PR adds a new method, but we already have suitable public methods, there is no need to make the public API more confusing by adding a new method that looks very similar to an existing method.

rserro:
Found the problem. Idle should have been set before broadcast. Now stop() works as expected, dismissing the device player.

rserro:
Hi @ryanheise ! Is this ok to be a fix in the next release?

whitecodel:
@rserro is right — we need to call `_broadcastState()` after updating `_justAudioEvent`, as it was causing issues. I've also raised a PR addressing the same problem. In the PR description, I’ve explained the changes and the reason behind them.

I was facing an issue where the notification would still show even after calling `dispose()`. After making this change, everything is working fine now.

You can merge either my PR or this one, whichever you prefer.

Please take a look --- @ryanheise — https://github.com/ryanheise/just_audio/pull/1456

ryanheise:
LGTM. I have now merged and published. @whitecodel apologies, I merged @rserro 's since it was submitted first, but I can also merge yours for credit if you think that there is justification for the slight ordering difference. I'd want to get it right, and can credit both if that's the case.

PR #1460, opened 2025-05-08
Fix play interrupted by load on web.

### LINKED ISSUE #1107: AbortError: The play() request was interrupted by a new load request.
In Chrome on the web version, when switching to a new song while the first song is still loading, I get the following error that interrupts the debug process:
`AbortError: The play() request was interrupted by a new load request. https://goo.gl/LdLk22`

https://github.com/mrandrey/just_audio_bug_report/

1. Launch in debug mode in Chrome
2. Quickly click on Song 1, Song 2, Song 3, Song 2, Song 1, etc.
3. The debug process should interrupt with the error.

**Error messages**

```
AbortError: The play() request was interrupted by a new load request. https://goo.gl/LdLk22
```

**Expected behavior**
I expect the loading/buffering to stop, then the new song to load.

 - OS: Mac OS Sonoma
 - Browser: Chrome

 - The problem is only on the web version.

**Flutter SDK version**
```
[✓] Flutter (Channel stable, 3.13.6, on macOS 14.0 23A344 darwin-x64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 14.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.3)
[✓] Connected device (2 available)
[✓] Network resources

• No issues found!
```

**Additional context**
The project is to have an album with preview songs and an ability to go to the "full view" on another screen to listen to the entire song if the user has purchased it. If the user clicks on preview, then quickly clicks on the full song button, then they'll get an error because the preview is still buffering while the full song on the next screen is already loading the next song. I need to stop the buffering/loading and load the next song. Ideally, I'd like to load only what the user clicks on, so I'd rather not use a playlist as the solution.

Thanks for looking into this!

ryanheise:
Fixes #1107

### DIFF
--- a/just_audio_web/CHANGELOG.md
+++ b/just_audio_web/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.16
+
+* Fix play interrupted by load.
+
 ## 0.4.15
 
 * Support errorCode, errorMessage.
--- a/just_audio_web/lib/just_audio_web.dart
+++ b/just_audio_web/lib/just_audio_web.dart
@@ -6,6 +6,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_web_plugins/flutter_web_plugins.dart';
 import 'package:just_audio_platform_interface/just_audio_platform_interface.dart';
+import 'package:synchronized/synchronized.dart' as synch;
 import 'package:web/web.dart';
 
 /// The web implementation of [JustAudioPlatform].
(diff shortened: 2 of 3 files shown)


Abu-Abdullah:
still facing this issue in chrome with 4.16

ryanheise:
I'm not surprised, given that nobody tested it. As mentioned in #1107 I have no way to reproduce this, and therefore no way to test a fix. If you have one, I would encourage you to open a new issue.

PR #1490, opened 2025-06-05
[LockCachingAudioSource] Fix cache renameSync delete exception

danielgomezriconeon:
Fix:
- https://github.com/ryanheise/just_audio/issues/1032


danielgomezriconeon:
changing to: https://github.com/ryanheise/just_audio/pull/1492

Issue #1475, opened 2025-05-22
MacOS visualizer branch build bug


w-rui:
**Which API doesn't behave as documented, and how does it misbehave?**

Podfile error:

```
  s.source_files = 'just_audio/Sources/just_audio/**/*.{h,m}'
```

Should be (`mm` not included):

```
  s.source_files = 'just_audio/Sources/just_audio/**/*.{h,m,mm}'
```

New added androidFFT.mm will not be compiled if `mm` not included.

**Minimal reproduction project**
"The example".

**To Reproduce (i.e. user steps, not code)**
Just build and run.

**Error messages**

```
[        ] Undefined symbols for architecture x86_64:
[        ]   "_OBJC_CLASS_$_AndroidFFT", referenced from:
[        ]        in AudioPlayer.o
[        ] ld: symbol(s) not found for architecture x86_64
[        ] clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

**Desktop (please complete the following information):**
 - OS: macos 14.7

**Flutter SDK version**
```
Flutter 3.32.1-0.0.pre.1 • channel [user-branch] • unknown source
Framework • revision 06523144fd (1 年 9 个月前) • 2023-08-22 16:58:01 +0800
Engine • revision 1881800949 (3 天前) • 2025-05-19 10:54:07 -0700
Tools • Dart 3.8.0 • DevTools 2.45.1
```

Mind-Reaper:
I have this issue also when trying to run on iOS

```
Error (Xcode): Undefined symbol: _OBJC_CLASS_$_AndroidFFT

Error (Xcode): Linker command failed with exit code 1 (use -v to see invocation)
```

ryanheise:
Closed via #1520

PR #1520, opened 2025-07-26
Modify just_audio.podspec to include `mm` in source files 

Mind-Reaper:
This fixes the #1475  issue with Android_FFT missing. Thanks to @w-rui


ryanheise:
Thanks for this. I remember this during the migration to SwiftPM, and then it slipped my mind, so your contribution is very helpful. It is now merged.

Issue #242, opened 2020-11-12
HLS support for Web


kcrebound:
**Is your feature request related to a problem? Please describe.**
Cant play HLS in the web

**Describe the solution you'd like**
Can't find info about the problem with HLS for web, any plans?

**Describe alternatives you've considered**
I can use it without HLS, but with huge audio files, it is a big problem.

ryanheise:
Note to self: this could be used: https://github.com/google/shaka-player/

liri2006:
There is a package which adds Web HLS support for video_player, maybe it can be used as inspiration for implementing HLS in just_audio - https://github.com/balvinderz/video_player_web_hls.

ryanheise:
Good find, @liri2006 . Internally it is using [hls.js](https://github.com/video-dev/hls.js/) instead of shaka and shaka also includes DASH support, but your link could be a useful guideline on how to integrate that into a Flutter plugin. Technically the javascript library can't directly be pulled in by the plugin, so there'll need to be some README instructions on what script tag the app developer will need to add to their app.

liorshk:
Hey @ryanheise ,
I managed to get this done here:
https://github.com/liorshk/just_audio/blob/704eb4de9d52dba41d751f378b4b56cabbefff31/just_audio_web/lib/just_audio_web.dart

Using hls.js library.

ryanheise:
Hi @liorshk , nice work! Sorry I missed this comment. Would you be interested in submitting a PR with instructions in the README for web setup instructions? Judging by the code, it also looks like the instructions could be optional for an app that doesn't actually use HLS.

PR #1521, opened 2025-07-27
Add support for HLS streams in just_audio_web

dindils:
Hi, thanks for maintaining this great library.

I needed HLS support for a project I'm working on so this change adds support for HLS streams in just_audio_web based on the work done by liorshk as discussed in issue [#242](https://github.com/ryanheise/just_audio/issues/242).

This utilizes HLS.JS for playback, added note in README.md on the script tag that needs to be added to fetch that library if HLS playback is desired.

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 0.10.5
 
-* Add support for HLS streams in just_audio_web (@dindils)
+* Add support for HLS streams in just_audio_web (@dindils, @liorshk, @aletorrado)
 
 ## 0.10.4
 
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.5
+
+* Add support for HLS streams in just_audio_web (@dindils)
+
 ## 0.10.4
 
 * Fix bug on simultaneous loads.
--- a/just_audio/README.md
+++ b/just_audio/README.md
@@ -352,7 +352,7 @@ For issues with the Linux implementation, please open an issue on the respective
 To enable HLS (HTTP Live Streaming) playback in the web player, add the following script tag to your HTML file:
 
 ```html
-<script src="//cdn.jsdelivr.net/npm/hls.js@1"></script>
+<script async src="//cdn.jsdelivr.net/npm/hls.js@1"></script>
 ```
 
 ## Troubleshooting
--- a/just_audio/README.md
+++ b/just_audio/README.md
@@ -347,6 +347,14 @@ dependencies:
 
 For issues with the Linux implementation, please open an issue on the respective implementation's GitHub issues page.
 
+### Web
+
+To enable HLS (HTTP Live Streaming) playback in the web player, add the following script tag to your HTML file:
+
+```html
+<script src="//cdn.jsdelivr.net/npm/hls.js@1"></script>
+```
+
 ## Troubleshooting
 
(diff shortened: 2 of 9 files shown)


ryanheise:
Thanks for this. Now first, you may (or may not) be aware that there is an existing PR #1246 by @aletorrado also using HLS.js, and so I would want to credit both of you in the CHANGELOG, although yours looks more up to date using extension types.

Can I get you to just get you to check #1246 and compare it with your own to see if it sparks any further changes to your own PR?

aletorrado:
Hi! Thanks @ryanheise for keeping my PR in mind.

I don't like the need to synchronously preload Hls.js, that would delay the initial load of the app. I'd prefer to lazy load Hls.js like in #1246. To avoid dependance on hard coded CDN's, I would just consider to self host the Hls.js bundle.

dindils:
@ryanheise Was not aware of that PR, thanks for pointing that out. No real changes that I see a need for except using async in the script tag. Updated the CHANGELOG to reflect the contributions for this feature.

@aletorrado Really good point! I updated the README.md to recommend to people to use the async tag when loading hls.js which I can see is a good improvement. Although I like your solution to load hls.js automatically when needed, I think I'd prefer to keep it as it is since then I don't have to worry about any hard-coded dependencies since it's up to the user of the library to provide it. For my use case I'm only streaming HLS files so I'd prefer to load it immediately anyways.

aletorrado:
I don't think just adding the async attribute will be the best solution, because that would introduce race conditions. Think about an application that plays a media file automatically on startup (maybe an app for a radio station), but Hls.js may not have loaded at that point. Somehow, the dart code should wait for Hls.js to load.

Regarding the URL for the Hls.js bundle, maybe it should be configurable, while defaulting to the official CDN provided by Hls.js.

@ryanheise what do you think about self hosting the js bundle for Hls.js?

ryanheise:
@aletorrado , I think as long as the app developer can choose between self hosting and CDN, it should be fine. So I rather than bundle a copy of hls.js with the plugin, the README could just give instructions.

What I'll do to get you both in the change log is I'll create a new `feature/hls` branch incorporating @dindils 's extension types, then @aletorrado would you be able to make a PR against that branch to merge in your proposed changes?

Edit: done. Let's continue on this branch #1523

@dindils I have reverted your "new formatter" changes since we're only on sdk ^3.3.0 (and it cleans up the diffs).

aletorrado:
@ryanheise Ok, just to check: Do you want me to integrate my Hls.js loader while preserving the integration into Html5AudioPlayer from @dindils fork?

ryanheise:
@aletorrado sorry for the delay (had an emergency to deal with).

> Ok, just to check: Do you want me to integrate my Hls.js loader while preserving the integration into Html5AudioPlayer from @dindils fork?

Yes, exactly. We include @dindils 's extension type code combined with your loading code and then (and I guess the README example script tag if it needs to change). That basically means you would create a new branch based on `feature/hls` as the starting point, and I can merge that. I see @dindils has already put you in the CHANGELOG so basically once that's merged I'd be happy to test and release.

dindils:
@aletorrado Any plans from you to work on this further or should we just try to get the HLS stream support in as it is? It would be nice for me to be able to track the mainline branch rather than have my custom patch for this support.

Also had found some additional things to fix in the HLS livestreaming support in just_audio_web but hadn't taken the time to integrate into this branch or a new pull request since I was unsure whether you had plans to work on this. See https://github.com/dindils/just_audio/commits/further-playback-fixes/ for more fixes.

Issue #1519, opened 2025-07-19
Audio output stops unexpectedly on Android 16.


FluffyBunniesTasteTheBest:
**Which API doesn't behave as documented, and how does it misbehave?**
`AudioPlayer.play()` is buggy on Android **16**:
- Audio output ends abruptly.
- Short audio clips sometimes don't start at all.
- Looping background noise sometimes stops when another `AudioPlayer` starts playing.

**Minimal reproduction project**
[just_audio_issue.zip](https://github.com/user-attachments/files/21328611/just_audio_issue.zip)

**To Reproduce (i.e. user steps, not code)**
1. Run the attached project on an Android 16 device like Google Pixel 9a (or 9 XL).
2. Push the buttons labeled "Toggle SFX Loop" and "Toggle Background Audio".
3. Listen to whats happening. Note: Sometimes after toggeling SFX loop and background audio a few times, it starts to work as intended. Simply restart the app to restore the faulty behavior. 

**Error messages**

```
Launching lib/main.dart on Pixel 9a in debug mode...
✓ Built build/app/outputs/flutter-apk/app-debug.apk
I/flutter (13423): [IMPORTANT:flutter/shell/platform/android/android_context_vk_impeller.cc(61)] Using the Impeller rendering backend (Vulkan).
I/ust_audio_issue(13423): Compiler allocated 5111KB to compile void android.view.ViewRootImpl.performTraversals()
Connecting to VM Service at ws://127.0.0.1:55735/vbjqplbWsIY=/ws
Connected to the VM Service.
I/Choreographer(13423): Skipped 31 frames!  The application may be doing too much work on its main thread.
I/ExoPlayerImpl(13423): Init c2e2150 [AndroidXMedia3/1.4.1] [tegu, Pixel 9a, Google, 36]
I/ust_audio_issue(13423): hiddenapi: Accessing hidden method Landroid/media/AudioTrack;->getLatency()I (runtime_flags=0, domain=platform, api=unsupported) from Landroidx/media3/exoplayer/audio/AudioTrackPositionTracker; (domain=app) using reflection: allowed
D/AudioSystem(13423): onNewServiceWithAdapter: media.audio_flinger service obtained 0xb40000766971d0e0
D/AudioSystem(13423): getService: checking for service media.audio_flinger: 0xb4000077796c6340
I/ExoPlayerImpl(13423): Init a20d62 [AndroidXMedia3/1.4.1] [tegu, Pixel 9a, Google, 36]
[...]

PR #1527, opened 2025-08-12
Add androidAudioOfflinePreferences parameter to AudioPlayer()

### LINKED ISSUE #1526: Allowing disabling the audio offload feature on Android completely (Speed change not working in some Android 15 devices, e.g. Samsung SM-S921B)
It's related to a potential issue in some Android OEM (e.g. some newer Samsung devices like S21 running on Android 15) not actually applying the playback speed in offload mode, despite specifying the `setIsSpeedChangeSupportRequired` when setting up the audio offload configuration. See more at https://github.com/androidx/media/issues/1101 and https://github.com/androidx/media/issues/1110 

A potential workaround is to expose an option to disable `Audio offload` completely on Android. 

I'm not sure if we can detect this when it happens. If we can actually can detect this, then perhaps a smarter way could be to only disable audio offload on android if we see that the speed is not actually set to what we want. However this might also affect other features like voice boosting, etc.

**Additional context**
We have received bug reports from only Samsung users with the exact behavior mentioned in the referenced Github issue after updating our Android target SDK to Android 15. However we couldn't reproduce this yet on our Samsung devices, I assume it's because we didn't have the latest kernel version on our older devices and this issue only occurs on newer phones.

Since the error is only logged as a warning (according to the referenced issue: `W  Failed to set playback params
java.lang.IllegalArgumentException: arguments out of range`), we didn't have that information in our production logs.

ryanheise:
Fixes #1526

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.10.5
+
+* Add androidAudioOffloadPreferences.
+* Deprecate androidOffloadSchedulingEnabled.
+
 ## 0.10.4
 
 * Fix bug on simultaneous loads.
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 0.10.5
 
+* Disable Android audio offload by default to prevent playback issues.
 * Add androidAudioOffloadPreferences.
 * Deprecate androidOffloadSchedulingEnabled.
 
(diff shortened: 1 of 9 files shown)


RisbourgNathan:
Hello, this pull request also seems to fix #1519 on my Android 16 Google Pixel 8 Pro when using the following parameters:
```
AudioPlayer(
  androidAudioOffloadPreferences: AndroidAudioOffloadPreferences(
    audioOffloadMode: AndroidAudioOffloadMode.disabled,
  ),
);
```

ryanheise:
Hmm, it is troubling if true that we can't rely on audio offload to be stable. Perhaps I ought to disable it by default after all.

ryanheise:
Confirmed, this does also resolve the Android 16 issue.

PR #1531, opened 2025-09-01
android https upgrade on the proxy server

renanyoy:
On Android, I had problems with clear Text http requests that I couldn't resolve by authorizing it in the manifest.

I tried to upgrade the proxy server in https, and it seems to work ;)

The https is only active on Android Platform to don't break other platforms.


renanyoy:
the certificates wasn't loading cause of addon assets path. I removed them totally, they even not be needed.

PR #1555, opened 2025-09-29
Fix bug by syncing ExoPlayer shuffleOrder on setShuffleOrder (Closes #1551)

### LINKED ISSUE #1551: Playlists: Auto-advance always uses the same effective sequence with shuffling enabled
After each call to the player's `shuffle()` method, it changes the effective sequence as intended.
`seekToPrevious()` and `seekToNext()` correctly follow the current effective sequence.
However, when we instead let the player auto-advance with shuffling enabled, the sequence it follows will stay the same across calls to `shuffle()`.
Presumably, auto-advance always uses the initial shuffleIndices permutation from before the first re-shuffle.

The example (example_playlist.dart)

1. Launch the playlist example: `flutter run -t lib/example_playlist.dart`
2. Enable shuffle mode by clicking on the shuffle icon.
3. Skip through the playlist starting from the first item by using the skip buttons, observe the playing order.
4. Go back to the first item (shuffle mode still enabled).
5. Seek to the end of the current item to trigger auto-advance. Repeat for the items that are being played after that.
6. Notice that the playing order this time was different from the order when using the skip buttons.
7. Re-shuffle an arbitrary amount of times by toggling shuffle mode off and on again, then repeat steps 4 and 5.
8. Notice that the playing order of auto-advance is still the same as before the re-shuffles.

If the playing order is not different when comparing the "skip button" approach with the "auto-advance" approach, it might be a coincidence, so re-shuffle and try again.

**Error messages**
None.

**Expected behavior**
I expected the effective playing sequences to be identical when comparing the "skip button" approach with the "auto-advance" approach.
I expected the effective playing sequence of the "auto-advance" approach to change after (some) re-shuffles.

**Screenshots**
None.

**Desktop:**
None.

**Smartphone 1:**
 - Device: Medium_Phone_API_36.0 (emulated)
 - OS: Android 16

**Smartphone 2:**
 - Device: Samsung Galaxy A41
 - OS: Android 12

**Flutter SDK version**
```
Flutter (Channel stable, 3.35.3, on macOS 15.6.1 24G90 darwin-arm64, locale en-DE)
```

**Additional context**
None.

LennartEnns:
This pull request solves Issue #1551.
**Warning**: Not thoroughly tested yet, but since effectively only 1 line has been changed, this should not take long.

### DIFF
--- a/just_audio/CHANGELOG.md
+++ b/just_audio/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.6
+
+* Fix bug where ExoPlayer's auto-advance ignores new shuffle order
+
 ## 0.10.5
 
 * Disable Android audio offload by default to prevent playback issues.
--- a/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
+++ b/just_audio/android/src/main/java/com/ryanheise/just_audio/AudioPlayer.java
@@ -583,11 +583,13 @@ public class AudioPlayer implements MethodCallHandler, Player.Listener, Metadata
         Map<?, ?> map = (Map<?, ?>)json;
         String id = mapGet(map, "id");
         MediaSource mediaSource = mediaSources.get(id);
+        ShuffleOrder shuffleOrder = decodeShuffleOrder(mapGet(map, "shuffleOrder"));
+        player.setShuffleOrder(shuffleOrder);
         if (mediaSource == null) return;
         switch ((String)mapGet(map, "type")) {
         case "concatenating":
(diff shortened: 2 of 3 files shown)


Chaphasilor:
We're able to reproduce this bug (#1551) in https://github.com/jmshrv/finamp/pull/1442.  
Thanks for whipping up the PR @LennartEnns, I've tested it and the issue is indeed fixed there.

There is another (potentially) related issue when toggling shuffle mode, where the `playbackEventStream` isn't emitting a new event even though the current index is changing due to shuffle being toggled.  
Edit: sorry for hitting send to early.

lachy874125:
This fixes the issue for me. Thanks for this!

PR #1574, opened 2026-02-10
Fix RangeError in shuffleIndices caused by shared mutable list reference

### LINKED ISSUE #1573: RangeError in shuffleIndices: shared mutable list reference between just_audio and just_audio_background

`ConcatenatingAudioSource.insert()`, `removeAt()`, and `removeRange()` cause a `RangeError` crash in `AudioSourceExtension.shuffleIndices` (in `just_audio_background`) when shuffle mode is active.

The root cause is `ConcatenatingAudioSource._toMessage()` passing `_shuffleOrder.indices` **by reference** (not a copy) into `ConcatenatingAudioSourceMessage`. This creates a shared mutable `List<int>` between the frontend (`just_audio`) and the background handler (`just_audio_background`).

When the frontend mutates `_shuffleOrder` during `insert()` or `removeAt()`, it immediately changes `_source.shuffleOrder` in the background handler while `_source.children` remains stale until the async `concatenatingInsertAll` request arrives — producing an off-by-one desync that triggers the crash.

Notably, `ConcatenatingInsertAllRequest` already copies correctly with `List.of(_shuffleOrder.indices)`, and `_toMessage()` also copies children via `.toList()` — only the `shuffleOrder` field in `_toMessage()` is missing the defensive copy.

The issue is timing-dependent and occurs when `insert()`/`removeAt()` is called on a `ConcatenatingAudioSource` while using `just_audio_background`, and `setShuffleMode()` or `_updateShuffleIndices()` is triggered during the `await` yield between the shuffle order mutation and the platform request. This is more likely on lower-end devices under load.

1. Set up a `ConcatenatingAudioSource` with `just_audio_background` and shuffle mode enabled
2. Rapidly add or remove items from the concatenating source (e.g. modifying a playlist)
3. During the async gap between the shuffle order mutation and the platform children update, if `_setPlatformActive` triggers `setShuffleMode` → `_updateShuffleIndices()`, the background handler reads a `shuffleOrder` that is already mutated but `children` that is still stale
4. `shuffleIndices` getter indexes `childIndicesList` with an out-of-bounds index → crash

**Error messages**

```
RangeError (length): Invalid value: Not in inclusive range 0..N: N+1
```

Stack trace points to `AudioSourceExtension.shuffleIndices` in `just_audio_background.dart` (~line 891-914).

**Expected behavior**

`insert()`, `removeAt()`, and `removeRange()` on `ConcatenatingAudioSource` should not cause a `RangeError` in the background handler, regardless of timing. The `shuffleOrder` and `children` should always be consistent from the background handler's perspective.

**Screenshots**

N/A — this is a runtime crash, not a visual issue.

N/A

- Device: Android (low-end devices, 2GB RAM)
- OS: Android 12+

**Flutter SDK version**

```
[✓] Flutter (Channel stable, 3.35.5, on macOS 26.2 25C56 darwin-arm64, locale en-CO)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2025.2)
[✓] Connected device (2 available)
[✓] Network resources
```

**Additional context**

**Versions:**
- just_audio: 0.10.5
- just_audio_background: 0.0.1-beta.17

**Suggested fix (two single-line changes):**

Fix 1 — Root cause (`just_audio.dart` ~line 3214): Copy the list in `_toMessage()`:
```dart
shuffleOrder: List.of(_shuffleOrder.indices),
```

Fix 2 — Defensive (`just_audio_background.dart` ~line 903): Bounds check in `shuffleIndices`:
```dart
for (final index in self.shuffleOrder) {
  if (index < childIndicesList.length) {
    indices.addAll(childIndicesList[index]);
  }
}
```

Both fixes are zero-risk and should be applied together. Fix 1 breaks the shared reference (root cause). Fix 2 adds a safety net for any remaining timing windows.

**Key observations:**
- `_setPlatformActive.setPlatform()` calls `setShuffleMode` BEFORE `_load`, widening the desync window
- Mutation methods hold their own `_lock` but don't coordinate with platform activation
- All other platform implementations serialize via `toMap()`, so reference identity is irrelevant — this only affects `just_audio_background`

danielgomezriconeon:
Fixes #1573

## Problem

`ConcatenatingAudioSource._toMessage()` passes `_shuffleOrder.indices` **by reference** into `ConcatenatingAudioSourceMessage`. This means the background handler (`just_audio_background`) and the frontend (`just_audio`) share the same `List<int>` object.

When `insert()` is called, it mutates `_shuffleOrder` synchronously — but the platform request to update `children` happens *after* an `await` yield. During that async gap, the background handler sees a `shuffleOrder` that already reflects the new size while `children` still has the old size. The `shuffleIndices` getter then indexes into `childIndicesList` with an out-of-bounds value, causing a `RangeError`.

Notably, `ConcatenatingInsertAllRequest` already copies correctly with `List.of(_shuffleOrder.indices)`, and `_toMessage()` also copies `children` via `.toList()` — only the `shuffleOrder` field was missing the defensive copy.

### Crashlytics stacktrace

```
RangeError (length): Invalid value: Not in inclusive range 0..N: N+1
```

Crash site: `just_audio_background.dart:896` → `AudioSourceExtension.shuffleIndices`

```dart
List<int> get shuffleIndices {
  final self = this;
  if (self is ConcatenatingAudioSourceMessage) {
    var offset = 0;
    final childIndicesList = <List<int>>[];
    for (final child in self.children) {        // N children → N entries
      childIndicesList.add(child.shuffleIndices.map((i) => i + offset).toList());
      offset += childIndices.length;
    }
    for (final index in self.shuffleOrder) {    // shuffleOrder has value N
      indices.addAll(childIndicesList[index]);   // 💥 childIndicesList[N] → RangeError
    }
  }
}
```

Two crash paths observed:

**Path A — Playlist load:**
```
setAudioSources → load → _setPlatformActive.setPlatform
  → setShuffleMode → _updateShuffleIndices → shuffleIndices → 💥
```

**Path B — Live queue mutation (e.g. inserting a track):**
```
ConcatenatingAudioSource.insert → _setPlatformActive.setPlatform
  → setShuffleMode → _updateShuffleIndices → shuffleIndices → 💥
```

The mismatch is exactly ±1 in every observed crash (e.g., `0..498:499`, `0..103:104`), matching exactly one item being inserted or removed — `children` count changes by 1 while `shuffleOrder` isn't atomically updated.

### Fix 1 — Root cause (`just_audio`)

Copy the list in `_toMessage()` so the background handler gets its own snapshot:

```dart
shuffleOrder: List.of(_shuffleOrder.indices)
```

This breaks the shared mutable reference. The background handler's `shuffleOrder` and `children` now stay consistent regardless of what the frontend does after the message is created.

### Fix 2 — Defensive bounds check (`just_audio_background`)

Add a guard in the `shuffleIndices` getter so it skips invalid indices instead of crashing:

```dart
for (final index in self.shuffleOrder) {
  if (index < childIndicesList.length) {
    indices.addAll(childIndicesList[index]);
  }
}
```

This is a safety net — Fix 1 prevents the desync, but this ensures `shuffleIndices` can never throw a `RangeError` even if a future code path reintroduces shared state.

## Tests

**`just_audio/test/just_audio_test.dart`** — 3 new tests:
- `_toMessage should copy shuffleOrder to prevent shared mutation` — verifies the message holds a defensive copy, not a shared reference
- `desync'd shuffleOrder causes RangeError in shuffleIndices computation` — proves the algorithm crashes when given desync'd input (the exact Crashlytics scenario)
- `insert should not cause shuffleOrder desync during async gap` — blocks `concatenatingInsertAll` mid-flight and verifies no desync occurs during the async gap

**`just_audio_background/test/shuffle_indices_test.dart`** — 3 new tests:
- `shuffleIndices returns correct indices for synced shuffleOrder` — normal case
- `shuffleIndices skips out-of-bounds indices in shuffleOrder` — verifies the bounds check
- `shuffleIndices handles all indices out of bounds` — edge case, all indices invalid

### DIFF
--- a/just_audio/lib/just_audio.dart
+++ b/just_audio/lib/just_audio.dart
@@ -3211,7 +3211,7 @@ class ConcatenatingAudioSource extends AudioSource {
       id: _id,
       children: children.map((child) => child._toMessage()).toList(),
       useLazyPreparation: useLazyPreparation,
-      shuffleOrder: _shuffleOrder.indices);
+      shuffleOrder: List.of(_shuffleOrder.indices));
 }
 
 /// An [AudioSource] that clips the audio of a [UriAudioSource] between a
--- a/just_audio/test/just_audio_test.dart
+++ b/just_audio/test/just_audio_test.dart
@@ -1056,6 +1056,103 @@ void runTests() {
     await player2.dispose();
   });
 
+  test('_toMessage should copy shuffleOrder to prevent shared mutation',
+      () async {
+    final source = ConcatenatingAudioSource(
+      shuffleOrder: DefaultShuffleOrder(random: Random(1001)),
+      children: [
+        AudioSource.uri(Uri.parse("https://foo.foo/a.mp3"), tag: 'a'),
+        AudioSource.uri(Uri.parse("https://foo.foo/b.mp3"), tag: 'b'),
+        AudioSource.uri(Uri.parse("https://foo.foo/c.mp3"), tag: 'c'),
+      ],
+    );
+    final player = AudioPlayer();
+    await player.setAudioSource(source);
+    final msg = mock.mostRecentPlayer!.loadedPlaylist!.children.first
+        as ConcatenatingAudioSourceMessage;
+
+    expect(msg.children.length, equals(3));
+    expect(msg.shuffleOrder.length, equals(3));
+
+    await source.insert(
+        3, AudioSource.uri(Uri.parse("https://foo.foo/d.mp3"), tag: 'd'));
+
+    expect(msg.shuffleOrder.length, equals(3),
+        reason: 'shuffleOrder should be a defensive copy, not mutated by insert');
+    expect(msg.children.length, equals(3));
+    expect(msg.shuffleOrder.length, equals(msg.children.length),
+        reason: 'shuffleOrder and children should stay in sync');
+
+    await player.dispose();
+  });
+
+  test(
+      'desync\'d shuffleOrder causes RangeError in shuffleIndices computation',
+      () {
+    final shuffleOrder = <int>[2, 0, 1];
+    final msg = ConcatenatingAudioSourceMessage(
+      id: 'test-cat',
+      children: [
+        ProgressiveAudioSourceMessage(
+            id: 'a', uri: 'https://foo/a.mp3', headers: null, tag: null),
+        ProgressiveAudioSourceMessage(
+            id: 'b', uri: 'https://foo/b.mp3', headers: null, tag: null),
+        ProgressiveAudioSourceMessage(
+            id: 'c', uri: 'https://foo/c.mp3', headers: null, tag: null),
+      ],
+      useLazyPreparation: true,
+      shuffleOrder: shuffleOrder,
+    );
+
+    expect(computeShuffleIndices(msg), equals([2, 0, 1]));
+
+    shuffleOrder.add(3);
+
+    expect(
+      () => computeShuffleIndices(msg),
+      throwsA(isA<RangeError>()),
+      reason:
+          'shuffleOrder references index 3 but children has only 3 entries',
+    );
+  });
+
+  test('insert should not cause shuffleOrder desync during async gap',
+      () async {
+    final source = ConcatenatingAudioSource(
+      shuffleOrder: DefaultShuffleOrder(random: Random(1001)),
+      children: [
+        AudioSource.uri(Uri.parse("https://foo.foo/a.mp3"), tag: 'a'),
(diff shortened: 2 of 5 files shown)


danielgomezriconeon:
@ryanheise friendly ping over here

ryanheise:
Hi @danielgomezriconeon Are your latest issues a repeat of this one or different?

danielgomezriconeon:
Sorry! I got confused about my work account and my personal one, sorry for the duplication, yes, those were the same problem

danielgomezriconeon:
@ryanheise friendly ping

